PR #2, opened 2019-03-05
Add config loader

roborourke:
Next steps will be adding module registration which will populate the default config before merging composer.json config settings.


load.php
```diff
@@ -0,0 +1,17 @@
+<?php
+/**
+ * Altis Core Loader.
+ *
+ * @package altis-core
+ */
+
+namespace Altis\Core;
```
joehoyle: For now we are sticking with the `HM\Platform` namesapce

load.php
```diff
@@ -0,0 +1,17 @@
...
+
+const ALTIS_CORE_DIR = __DIR__;
+
+include ALTIS_CORE_DIR . '/inc/config.php';
+
+// Define the env type as local if not set higher up the chain.
```
joehoyle: I wonder if there's a better way to handle the `HM_ENV_TYPE` default?

inc/config.php
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Retrieve the configuration for HM Platform.
+ *
+ * The configuration is defined by merging the defaults with the various files that allow to customise a particular
+ * installation.
+ *
+ * @return array Configuration data.
+ */
+function get_config() {
```
joehoyle: Wondering if we need this to be behind a class, so it gets autoloaded. Ala `HM\Platform\Config::get()` or something?

inc/config.php
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/**
+ * Return the value for a particular setting from the configuration,
+ *
+ * @param string $key Settings key to retrieve the value from.
+ *
+ * @return mixed Settings value.
+ */
+function get_config_value( string $key ) {
```
joehoyle: I wonder if it's worth adding a `default` arg? I think triggering an error on not found is a bit harsh to the calling code too!

inc/config.php
```diff
@@ -0,0 +1,146 @@
...
+
+	return $config[ $key ];
+}
+
+/**
+ * Merge the defaults and the contents of the various configuration files into a single configuration.
+ *
+ * @return array Configuration data.
+ */
+function get_merged_defaults_and_customisations() : array {
```
joehoyle: `customizations` I think, American spelling for code.

inc/config.php
```diff
@@ -0,0 +1,146 @@
...
+	$contents = json_decode( file_get_contents( $file ), true );
+
+	if ( ! is_array( $contents ) ) {
+		trigger_error( 'Decoding the JSON in ' . $file . ' .', E_USER_WARNING );
+		return [];
+	}
+
+	// Get the .extra.altis settings.
+	if ( ! isset( $contents['extra'], $contents['extra']['altis'] ) ) {
+		trigger_error( 'No altis configuration found in composer.json under extra.altis', E_USER_WARNING );
```
joehoyle: I don't know that we want to trigger this. If you don't have an altis section, I think that should be ok?

inc/config.php
```diff
@@ -0,0 +1,146 @@
...
+		return [];
+	}
+
+	return $contents['extra']['altis'];
```
joehoyle: See below for naming, I think for now we stay with "platform" and then do a full renaming once it's confirmed.

inc/config.php
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/**
+ * Get the contents of a JSON file, decode it, and return as an array.
+ *
+ * @param string $file Path to the JSON file.
+ *
+ * @return array Decoded data in array form, empty array if JSON data could not read.
+ */
+function get_json_file_contents_as_array( $file ) {
```
joehoyle: This could be `( $file ) : array`

inc/config.php
roborourke: To be honest I don't think this got used anywhere in the end on the `beta` branch, going to remove it for now

inc/config.php
```diff
@@ -0,0 +1,146 @@
...
+	// Look for a `composer.json` file.
+	if ( is_readable( $composer_file ) ) {
+		$config = get_merged_settings( $config, get_json_file_contents_as_array( $composer_file ) );
+	} else {
+		trigger_error( 'A composer file could not be found at ' . $composer_file, E_USER_WARNING );
+		return $config;
+	}
+
+	// Look for environment specific settings in the config and merge.
+	if ( defined( 'HM_ENV_TYPE' ) && isset( $config['environments'], $config['environments'][ HM_ENV_TYPE ] ) ) {
```
joehoyle: I added `HM\Platform\get_environment_type()` in #1

load.php
roborourke: It's done in the `hm-platform/load.php` file in the beta branch so this is pretty much copy paste from there. I think it should happen at a low level but perhaps in the wp-config.php in the CMS module? Needs to happen before the config is called anywhere.

inc/config.php
roborourke: Yep, switching to classmap.

inc/class-config.php
```diff
@@ -0,0 +1,163 @@
...
+
+	/**
+	 * The environment type is used to override the config
+	 * as well as allow for checks within a project codebase.
+	 *
+	 * Defaults to local, this constant will be set at the stack level.
+	 *
+	 * @return string The environment type string.
+	 */
+	static function get_environment_type() : string {
```
joehoyle: I think we might need to generally fix this pattern of reliance on functions. For anyone else: this has been copied from `inc/namespace.php` because there's no guerentee that function would be available yet here.

Either we need to accept `Config::get()` can only be used after the autoloader has run (defeating the purpose of using a class for Config`, or we require no functions be used from elsewhere.

Any ideas @rmccue ?

inc/class-config.php
```diff
@@ -0,0 +1,163 @@
...
+	 *
+	 * @param array $config    Existing configuration.
+	 * @param array $overrides Settings to merge in.
+	 *
+	 * @return array Configuration data.
+	 */
+	protected static function get_merged_settings( array $config, array $overrides ) : array {
+		$merged = $config;
+
+		foreach ( $overrides as $key => &$value ) {
```
joehoyle: Why is `$value` passed by ref here?

inc/class-config.php
```diff
@@ -0,0 +1,163 @@
...
+			if ( is_array( $value ) && isset( $merged[ $key ] ) && is_array( $merged[ $key ] ) ) {
+				$merged[ $key ] = self::get_merged_settings( $merged[ $key ], $value );
+			} elseif ( is_numeric( $key ) ) {
```
joehoyle: I'm not 100% sure of what this is doing, is it merging arrays? If so, do we want that? What if I want to override an array?

inc/class-config.php
```diff
@@ -0,0 +1,163 @@
...
+	 * with any overrides present in composer.json.
+	 *
+	 * @return array Configuration data.
+	 */
+	static function get() : array {
+		if ( self::$config ) {
+			return self::$config;
+		}
+
+		self::$config = self::get_merged_defaults_and_customizations();
```
joehoyle: Thinking we should pass this through a filter so anyone to make modifications to.

inc/class-config.php
```diff
@@ -0,0 +1,163 @@
...
+			'modules' => [],
+			'environments' => [
+				'local' => [],
+				'development' => [],
+				'staging' => [],
+				'production' => [],
+			],
+		];
+
+		// @todo get registered modules & their defaults here.
```
joehoyle: Do we need to process module defaults? I'm wondering if that could be do at another level, probably mostly in the module themselves. I could see it also working here, but just thinking we probably don't want a `@todo` for it, but we can add it if we need it at some point.

inc/class-config.php
```diff
@@ -0,0 +1,163 @@
...
+
+		if ( ! defined( 'ROOT_DIR' ) ) {
```
joehoyle: We should set this in the `wp-config.php` of CMS probably, so we don't need to check for it.

inc/class-config.php
roborourke: It's probably a bit of an anti-pattern but the helper functions could be included in the `Config::get()` method, or included at the top of the config class file. Both of those options are a bit gross. Would likely need to include the helpers file in the modules object too so it's available in module bootstrap code.

inc/class-config.php
roborourke: I just found this on php.net for recursively merging arrays - forgot the ref is default behaviour now

inc/class-config.php
roborourke: Yeah basically deep merging arrays. This provides more flexibility. If we want to override an array we'll need to be very specific about the expected schema, This was a bit of laziness on my part. So lets say we could only merge at the top level eg, activation and settings, so any settings values even they're an array are fully overwritten

inc/class-config.php
roborourke: I was thinking of filtering late when the config actually gets used but can see future use cases for this 👍

inc/class-config.php
roborourke: So it depends if you're expecting to get the full config or not which is extended by modules as they register themselves with any default activation state or settings.

inc/class-config.php
roborourke: Added that to the `cms` repo PR 👍

inc/class-config.php
joehoyle: I think we could, but we should just add that if / when the module default settings comes into it. I was mostly treating the Config as arbitrary for the time being.

inc/class-config.php
roborourke: Gone for a more specific approach, we'll need to extend it as we come up with extra features for the config.

inc/class-config.php
roborourke: Resolved, will be using `load.php` and actions/filters.

inc/namespace.php
```diff
@@ -1,9 +1,115 @@
...
+
+	if ( ! is_readable( $file ) ) {
+		// phpcs:ignore
+		trigger_error( 'Could not read ' . $file . ' file.', E_USER_WARNING );
+		return [];
+	}
+
+	$contents = json_decode( file_get_contents( $file ), true );
+
+	if ( ! is_array( $contents ) ) {
```
rmccue: Should use `json_last_error() !== JSON_ERROR_NONE` instead, and maybe `json_last_error_msg()` for the message.

inc/namespace.php
```diff
@@ -1,9 +1,115 @@
...
+function merge_config_settings( array $config, array $overrides ) : array {
+	foreach ( $overrides as $key => $value ) {
+		switch ( $key ) {
+			// Merge module settings together.
+			case 'modules':
+				foreach ( $value as $module => $settings ) {
+					$config[ $key ][ $module ] = array_merge( $config[ $key ][ $module ] ?? [], $settings );
+				}
+				break;
+			// Replace property by default.
```
rmccue: Newline before this line.

inc/namespace.php
```diff
@@ -1,9 +1,115 @@
...
+	$config = merge_config_settings( $config, $config['environments'][ $environment ] ?? [] );
+
+	return $config;
+}
+
+/**
+ * Merge settings in an existing configuration file.
+ *
+ * @param array $config    Existing configuration.
+ * @param array $overrides Settings to merge in.
```
rmccue: Shouldn't be aligned.

inc/namespace.php
```diff
@@ -1,9 +1,115 @@
...
+	if ( empty( $config ) ) {
+		$config = get_merged_config();
+	}
+
+	/**
+	 * Filter the entire platform config.
+	 *
+	 * @param array $config The full config array.
+	 */
+	$config = apply_filters( 'hm-platform.config', $config );
```
rmccue: This is inconsistent on repeated calls, as the second time around it will re-filter the previously-filtered code, so might break. Imagine something like:

```
add_filter( 'hm-platform.config', function ( $data ) {
    $data['foo'] .= 'blah';
    return $data;
} );
```

PR #3, opened 2019-03-07
Add module object and loading

roborourke:


inc/class-module.php
```diff
@@ -0,0 +1,148 @@
...
+	 * @var array
+	 */
+	protected static $modules = [];
+
+	/**
+	 * The slug for the module, used to reference individual modules.
+	 *
+	 * @var string
+	 */
+	protected $slug;
```
rmccue: I think we need an accessor for `slug` as well?

inc/class-module.php
```diff
@@ -0,0 +1,148 @@
...
+ * The Module object provides a common interface for registering
+ * modules and carrying out module related tasks.
+ *
+ * @package hm-platform
+ */
+
+namespace HM\Platform;
+
+class Module {
+
```
rmccue: Bonus newline.

inc/namespace.php
```diff
@@ -185,3 +185,40 @@ function fix_plugins_url( string $url, string $path, string $plugin ) : string {
...
+
+	return $enabled;
+}
+
+/**
+ * Load all enabled plugins, along with their customisation files.
+ */
+function load_enabled_modules() {
+	foreach ( get_enabled_modules() as $name => $module ) {
+
```
rmccue: Bonus newline.

load.php
```diff
@@ -3,4 +3,13 @@
  * Main entry point loader for the Core module.
  */
 
+// Get module functions.
+include './inc/namespace.php';
```
rmccue: `__DIR__` instead of `.` so it isn't affected by the cwd

load.php
```diff
@@ -3,4 +3,13 @@
...
+
+// Patch plugins URL for vendor directory.
 add_filter( 'plugins_url', 'HM\\Platform\\fix_plugins_url', 10, 3 );
+
+// Load modules.
+add_action( 'hm-platform.loaded_autoloader', function () {
+	HM\Platform\load_enabled_modules();
```
rmccue: No need for the anonymous func here?

load.php
rmccue: This should be `require_once` too

inc/class-module.php
roborourke: Sure, I'd left it out because that's the array key/accessor, so presumably you'd have the slug already.

inc/class-module.php
```diff
@@ -0,0 +1,155 @@
...
+	 * @var string
+	 */
+	protected $title;
+
+	/**
+	 * Settings array.
+	 *
+	 * @var array
+	 */
+	protected $settings;
```
rmccue: Need to document exactly what "settings" means here; it's essentially configuration for the module _itself_

inc/class-module.php
```diff
@@ -0,0 +1,155 @@
...
+
+	/**
+	 * Calls the module loader function.
+	 *
+	 * @return void
+	 */
+	public function load() {
+		if ( $this->settings['enabled'] ?? false ) {
+			$this->loader( $this->settings );
+		}
```
rmccue: We discussed getting rid of this method and having `::register` hook the loader into the action instead

inc/class-module.php
```diff
@@ -0,0 +1,155 @@
...
+		return $this->settings;
+	}
+
+	/**
+	 * Get a module setting by name.
+	 *
+	 * @return mixed
+	 */
+	public function get_setting( $name ) {
+		return $this->settings[ $name ] ?? false;
```
joehoyle: This should probably return `null` rather than `false` if it's not set.

inc/class-module.php
```diff
@@ -0,0 +1,155 @@
...
+	 * Registers a module with the store.
+	 *
+	 * @param string $slug
+	 * @param string $directory
+	 * @param string $title
+	 * @param array $settings
+	 * @param callable $loader
+	 * @return Module
+	 */
+	public static function register( string $slug, string $directory, string $title, array $settings, callable $loader ) : Module {
```
joehoyle: Is `settings` everywhere in this actually "defualt_settings"? Wondering at what stage the config settings get merged in, or if they do at all.

If I want to get a setting for a module, do I go via the `Module` object, or via the `Config` object? Assuming the latter, in which case maybe it makes more sense to call it `default_settings`?

_Or_, is `$settings` meant to be more like the schema for the possible settings for that module?

joehoyle: I'm thinking maybe it would be good to have a `register_module` function, so the `Module` class is purely an implementation detail that the caller never has to worry about.

inc/class-module.php
```diff
@@ -38,43 +38,52 @@ class Module {
...
-		$module = new Module( $slug, $directory, $title, $settings, $loader );
+	public static function register( string $slug, string $directory, string $title, ?array $settings = null, ?callable $loader = null ) : Module {
+		$module = new Module( $slug, $directory, $title, $settings );
 
 		// Store the module.
 		self::$modules[ $slug ] = $module;
 
+		// Add the loader to the module's loaded action.
+		if ( is_callable( $loader ) ) {
+			add_action( "hm-platform.modules.{$slug}.loaded", $loader );
```
joehoyle: Maybe we should hook this on `1`? As really I would expected the module to always be loaded on this hook.

inc/class-module.php
roborourke: Added this to `inc/namespace.php`.

roborourke: Yeah I think it'll be clearer to rename this param `$default_settings` 👍 

The final settings will indeed come from the config array, something I was doing previously but not implemented here as it should be. Will update to fix that.

load.php
```diff
@@ -3,4 +3,28 @@
...
+namespace HM\Platform;
+
+// Get module functions.
+require_once __DIR__ . '/inc/namespace.php';
+
+// Patch plugins URL for vendor directory.
 add_filter( 'plugins_url', 'HM\\Platform\\fix_plugins_url', 10, 3 );
+
+// Fire module init hook and load enabled modules.
+add_action( 'hm-platform.loaded_autoloader', function () {
```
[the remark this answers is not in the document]
roborourke: Should we have an RFC for module structure? Not sure if w formalised any best practices there. We could perhaps do that after MVP once we've ironed out the kinks and edge cases.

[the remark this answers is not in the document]
roborourke: We can refactor later but feel that it's clearer what's happening with it this way now.

PR #5, opened 2019-03-19
Add SEO module packages to install list

roborourke:
Current deps for SEO module

### DIFF
--- a/inc/composer/class-installer.php
+++ b/inc/composer/class-installer.php
@@ -23,6 +23,11 @@ class Installer extends BaseInstaller {
 			'humanmade/require-login',
 			'humanmade/stream',
 			'10up/elasticpress',
+			'humanmade/hm-redirects',
+			'humanmade/msm-sitemap',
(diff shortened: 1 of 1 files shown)


joehoyle:
Are the `wppackagist` ones available on packagist? I thought that was it's own repo -- and we can't use other repos, as `repositories` is a root-only composer field.

roborourke:
@joehoyle hm, so this is only working because my root composer repo specifies wpackagist then... that sucks. I'll try and experiment. We could potentially bake that requirement into our composer plugin, or add wpackagist programatically through it.

joehoyle:
@roborourke yeah I tried that yesterday, turns out plugins can not add repositories, as on a fresh install, the repo configuration and querying is done before installing plugins, so there's no way to get a repo added other than in the root composer config. In cases where I've needed a WP plugin, I've had to fork the repo, change the composer namespace and submit it to packagist under the humanmade namespace.

PR #13, opened 2019-04-05
Add package for analytics module

roborourke:


roborourke:
@rmccue this one can go too now

PR #16, opened 2019-04-16
Add cmb2 to exclude list

roborourke:
Required for humanmade/platform-cms#32


joehoyle:
Would recommend we wait for humanmade/platform-cms#32 to be merged before mergin this

roborourke:
Closing out as we’re not going to go ahead with this any time soon

PR #27, opened 2019-05-16
Register composer installer on init

joehoyle:
Composer installers take precedence by latest-added. Out installer can get
added before the composer-installers installer. That means that it will be
responsible for installing wordpress-plugin packages, which means we don't
get to install them into `vendor` based off our whitelist.

We instead register the installer "late" on `init`, which happens after all
plugins are loaded, but before packages are installed.

### DIFF
--- a/inc/composer/class-installer.php
+++ b/inc/composer/class-installer.php
@@ -14,7 +14,7 @@ class Installer extends BaseInstaller {
 	 * @return bool
 	 */
 	public function supports( $type ) {
-		return $type === 'wordpress-plugin';
+		return in_array( $type, [ 'wordpress-plugin', 'wordpress-muplugin' ], true );
 	}
 
 	public function getInstallPath( PackageInterface $package, $framework_type = '' ) {
--- a/inc/composer/class-installer.php
+++ b/inc/composer/class-installer.php
@@ -6,8 +6,18 @@ use Composer\Installers\Installer as BaseInstaller;
 use Composer\Package\PackageInterface;
 
 class Installer extends BaseInstaller {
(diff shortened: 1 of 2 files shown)


inc/composer/class-installer.php
```diff
@@ -6,8 +6,18 @@
...
-	public function getInstallPath( PackageInterface $package, $framework_type = '' ) {
 
+	/**
+	 * Check if the installer supports a given type.
+	 *
+	 * @param string $type
+	 * @return bool
+	 */
+	public function supports( $type ) {
+		return in_array( $type, [ 'wordpress-plugin', 'wordpress-muplugin' ], true );
```
roborourke: Do we need `wordpress-dropin` in this list too?

inc/composer/class-plugin.php
```diff
@@ -15,14 +16,34 @@ class Plugin implements PluginInterface {
...
+		];
+	}
+
+	/**
+	 * Register the installer on the `init` hook.
+	 *
+	 * We want to register later than the composer-installers installer
+	 * as the last-added installer takes precedence. This event only runs
+	 * on update (if this plugin is already present), so we have to add it
+	 * in addition ot the $this->activate() method.
```
roborourke:
```suggestion
	 * in addition to the $this->activate() method.
```

inc/composer/class-installer.php
joehoyle: We don't currently specify it in https://github.com/humanmade/altis-core/blob/master/inc/composer/class-installer.php#L42 so we haven't needed it up to this point.

PR #33, opened 2019-05-23
Add about pages

### LINKED ISSUE #32: Add an about page in the admin
Should contain:

- Version
- Module versions
- License information
- Stack specs
- Docs / resources links

This could potentially show some information we previously had on the cloud page UI such as latest deployment but not a hard requirement right now.

rmccue:
Fixes #32. Needs a lot more stuff done first though. We should also get design and marketing passes here, as right now all I've done is spit out all the raw data.

## About Page

Screenshot_2019-05-23 About ‹ WordPress Site — Altis

## Credits Page

Screenshot_2019-05-23 Credits ‹ WordPress Site — Altis(1)

### DIFF
--- a/inc/about/namespace.php
+++ b/inc/about/namespace.php
@@ -1,0 +1,220 @@
+<?php
+
+namespace Altis\About;
+
+use function ALtis\get_composer_data;
+use Altis\Module;
+use WP_Admin_Bar;
+
+const PAGE_SLUG = 'altis-about';
+
+/**
+ * Bootstrap About pages.
+ */
+function bootstrap() {
+	add_action( 'admin_bar_menu', __NAMESPACE__ . '\\register_menu_item', 0 );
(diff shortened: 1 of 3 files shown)


joehoyle:
@rmccue for credits, I added the authors, as actually the package namespace doesn't necessarily give credits as it might be under the hmn namespace. Thoughts on addin a new col to the table ala:

rmccue:
I probably wouldn't bother; I don't think we need it. The links and namespace are fine for attribution I think.

(Also, everything under the `humanmade` namespace likely has the IP owned by HM anyway, regardless of authorship.)

joehoyle:
> (Also, everything under the humanmade namespace likely has the IP owned by HM anyway, regardless of authorship.)

I was mentioning this specifically because we publish some packages of 3rd parties (like stream, elasticpress etc) to packagist under the human made url.

PR #36, opened 2019-05-31
Add autoloader class

### LINKED ISSUE #35: Add autoloader helper
We should port across the [autoloader helper](https://github.com/humanmade/hm-base/blob/master/content/mu-plugins/class-autoloader.php), as it significantly improves usability when developing. The alternative currently is to use classmap autoloaders via Composer, which requires regenerating constantly with Composer, and also requires you to do more setup.

rmccue:
Fixes #35.

Still to do:
* [x] Add documentation on how to use
* [x] Add documentation on _when_ to use

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -30,6 +30,25 @@ Returns an instance of the base AWS SDK with preconfigured credentials.
 
 The credentials can be supplied locally by defining the constants `HM_ENV_REGION`, `AWS_KEY` and `AWS_SECRET`.
 
+## Autoloader
+
+For convenience, an autoloader which follows the WordPress class file naming standard is available.
+
+**`register_class_path( string $prefix, string $path ) : void`**
+
+Registers the given prefix to be autoloaded from the given path. The prefix can either be a legacy-style class prefix (e.g. `Altis_CMS_`) or a namespace (e.g. `Altis\CMS`).
+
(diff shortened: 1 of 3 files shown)


docs/README.md
```diff
@@ -30,6 +30,25 @@ Returns an instance of the base AWS SDK with preconfigured credentials.
...
+
+Classes are mapped to paths through the following process:
+
+1. The prefix is removed from the class name
+2. The class name is lower-cased
+3. If the class name contains namespace separators, these are converted to directory separators
+4. Underscores in the class name (but not the parent namespace) are converted to dashes
+5. The name is prefixed with `class-`
+
+For example, with `register_class_path( 'Altis\CMS', '/dir' )`, the class `Altis\CMS\Alpha_Beta\Gamma_Delta` would be expected to live at `/dir/alpha_beta/class-gamma-delta.php`.
```
roborourke: Does this work with both `'Altis\CMS'` and `'Altis\\CMS'`?

Also I know it's part of our coding standards but would love to have dashes in directory names unless there's a good reason not to. The inconsistency bothers me 😛

rmccue: Yeah, `'Altis\CMS'` and `'Altis\\CMS'` are the same string in PHP, but the latter with a redundant escape.

PR #43, opened 2019-07-18
Add helper function get_environment_codebase_revision

### LINKED ISSUE #34: Helper function for code revision
There's no helper function for getting the current deployed code revision, which used to be available via `HM_APPLICATION_REVISION`. Rather than the constant rename, we've moved to using functions so we should add one for that.

It _might_ make more sense to have this in the Cloud module, not sure!

joehoyle:
Fixes #34.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -194,6 +194,27 @@ function get_environment_type() : string {
 }
 
 /**
+ * Get the current revision of the codebase deployed to the current environment.
+ *
+ * @return string
+ */
+function get_environment_codebase_revision() : string {
+	// This is the constant that is defined in wp-config-production.php
+	// in the EC2 architecture in Altis Cloud.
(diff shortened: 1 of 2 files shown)


inc/namespace.php
```diff
@@ -193,6 +193,27 @@ function get_environment_type() : string {
...
+		return HM_APPLICATION_REVISION;
+	}
+
+	// This is the constant that is defined in wp-config-production.php
+	// in the ECS architecture in Altis Cloud.
+	if ( defined( 'HM_DEPLOYMENT_REVISION' ) ) {
+		return HM_DEPLOYMENT_REVISION;
+	}
+
+	return 'unknown';
```
roborourke: Thinking of the places where this would be used - most likely for setting the version string when enqueuing scripts & styles, therefore I think it should fall back to returning `null`.

PR #51, opened 2019-08-21
Alphabetize excluded plugins list

### LINKED ISSUE #50: Alphabetize plugin override list
This list [`inc/composer/class-installer.php#L26`](https://github.com/humanmade/altis-core/blob/master/inc/composer/class-installer.php#L26) of plugins we exclude from normal type-based path determination is not alphabetized, and organizing it could improve ease-of-maintenance going forward.

kadamwhite:
This should make it easier to maintain and avoid unwanted plugin duplication in the future

Fixes #50


kadamwhite:
Not sure how there could be, none of our usages of this array appear to account for order. Thanks for the review!

kadamwhite:
(I should note that local testing and installation has worked as expected, so I'm not completely YOLO'ing this)

PR #57, opened 2019-10-16
Add simple local avatars to vendor

pdewouters:


rmccue:
This is for https://github.com/humanmade/altis-cms/pull/111

roborourke:
This is no longer necessary after the new installer plugin was added.

PR #64, opened 2020-01-28
add php-basic-auth to list of excluded plugins

jazzsequence:
allow it to be installed to the vendor directory rather than content/mu-plugins

required for https://github.com/humanmade/altis-security/issues/45


rmccue:
@roborourke do we want to backport to v3?

roborourke:
@rmccue no it's a new feature, unless there's demand for it in the short term I'd rather keep the number of RC releases down, they're a pain to maintain!

PR #67, opened 2020-02-05
Fix module settings to array if boolean set in config

### LINKED ISSUE #63: Coerce module settings to array if boolean set in config
If a module configuration is set to a boolean value eg. `false` to turn it off there's a merge error and a warning thrown.

We need to check the type of the value of `$settings` in the following lines:

https://github.com/humanmade/altis-core/blob/f0d4e1a1230a149d6c84c99756657cbf7bdcdd71/inc/namespace.php#L82-L85

If the type is a boolean then the value should be set to:

```php
$settings = [
    'enabled' => settings
];
```

If the type is anything other than a boolean or array then we should trigger a warning using `trigger_error( ..., E_USER_WARNING )` indicating which module settings are broken and then `continue;` the loop.

Acceptance criteria:

- [ ] Setting a module in `composer.json` config to `true` or `false` should not trigger an `array_merge` warning
- [ ] A boolean value set for a module in `composer.json` should be used as the `enabled` setting value

-----------

## Original iisue

Similar to [this issue](https://github.com/humanmade/altis-local-chassis/issues/67), I'm seeing this error from the site's PHP logs:

```
PHP Warning:  array_merge(): Argument #2 is not an array in /chassis/vendor/altis/core/inc/namespace.php on line 83
```

Here's the snippet of the project's composer config:

```json
{
  "extra": {
     "altis": {
      "modules": {
        "media": {
          "smart-media": false,
          "tachyon": false
        },
        "workflow": false,
        "seo": {
          "redirects": false
        },
        "analytics": {
          "native": {
            "experiments": {
              "titles": false
            }
          }
        }
      }
    }
  }
}
```

dashaluna:
Fixes #63 

@roborourke I've struggle to test this :( Changing values in the `composer.json` and then looking at `php_errors.log` felt a bit temperamental and still displaying the error when I "think" it should have worked.

I'd hugely appreciate if you could test it and double check. Or let me know how I can correctly test this. Thank you!

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -80,7 +80,23 @@ function merge_config_settings( array $config, array $overrides ) : array {
 			// Merge module settings together.
 			case 'modules':
 				foreach ( $value as $module => $settings ) {
-					$config[ $key ][ $module ] = array_merge( $config[ $key ][ $module ] ?? [], $settings );
+					// Check config settings syntax is valid.
+					$config_setting = $config[ $key ][ $module ];
+
+					if ( ! is_array( $config_setting ) && ! is_bool( $config_setting ) ) {
+						// phpcs:ignore
+						trigger_error( "Settings for the module '{$module}' specified in the composer.json are incorrect. It should be either an object or a boolean.", E_USER_WARNING );
+						continue;
+					}
+
+					// Convert bool value into an array with `enabled` setting.
+					if ( is_bool( $config_setting ) ) {
+						$config_setting = [
+							'enabled' => $config_setting,
+						];
+					}
+
+					$config[ $key ][ $module ] = array_merge( $config_setting, $settings );
 				}
 				break;
 
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -80,23 +80,22 @@ function merge_config_settings( array $config, array $overrides ) : array {
 			// Merge module settings together.
 			case 'modules':
 				foreach ( $value as $module => $settings ) {
-					// Check config settings syntax is valid.
-					$config_setting = $config[ $key ][ $module ];
+					$config_settings = $config[ $key ][ $module ] ?? [];
 
-					if ( ! is_array( $config_setting ) && ! is_bool( $config_setting ) ) {
-						// phpcs:ignore
+					// Check config settings syntax is valid.
+					if ( ! is_array( $config_settings ) && ! is_bool( $config_settings ) ) {
 						trigger_error( "Settings for the module '{$module}' specified in the composer.json are incorrect. It should be either an object or a boolean.", E_USER_WARNING );
 						continue;
 					}
 
 					// Convert bool value into an array with `enabled` setting.
-					if ( is_bool( $config_setting ) ) {
-						$config_setting = [
-							'enabled' => $config_setting,
+					if ( is_bool( $config_settings ) ) {
+						$config_settings = [
+							'enabled' => $config_settings,
 						];
 					}
 
-					$config[ $key ][ $module ] = array_merge( $config_setting, $settings );
+					$config[ $key ][ $module ] = array_merge( $config_settings, $settings );
 				}
 				break;
 
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -80,22 +80,21 @@ function merge_config_settings( array $config, array $overrides ) : array {
 			// Merge module settings together.
 			case 'modules':
 				foreach ( $value as $module => $settings ) {
-					$config_settings = $config[ $key ][ $module ] ?? [];
 
-					// Check config settings syntax is valid.
-					if ( ! is_array( $config_settings ) && ! is_bool( $config_settings ) ) {
+					// Check settings syntax is valid.
+					if ( ! is_array( $settings ) && ! is_bool( $settings ) ) {
 						trigger_error( "Settings for the module '{$module}' specified in the composer.json are incorrect. It should be either an object or a boolean.", E_USER_WARNING );
 						continue;
 					}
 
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -80,7 +80,23 @@ function merge_config_settings( array $config, array $overrides ) : array {
 			// Merge module settings together.
 			case 'modules':
 				foreach ( $value as $module => $settings ) {
-					$config[ $key ][ $module ] = array_merge( $config[ $key ][ $module ] ?? [], $settings );
+					// Check config settings syntax is valid.
+					$config_setting = $config[ $key ][ $module ];
+
+					if ( ! is_array( $config_setting ) && ! is_bool( $config_setting ) ) {
+						// phpcs:ignore
```
roborourke: Was this here because of local linter errors or hm-linter? We use this pattern all over the place so you shouldn't need it. Might need to check you have the latest HM coding standards installed.

If it is needed by hm-linter still I usually like to specify the sniff it's failing on eg `// phpcs:ignore Name.Of.Rule` so that it's only ignoring that rule.

inc/namespace.php
```diff
@@ -80,7 +80,23 @@ function merge_config_settings( array $config, array $overrides ) : array {
 			// Merge module settings together.
 			case 'modules':
 				foreach ( $value as $module => $settings ) {
-					$config[ $key ][ $module ] = array_merge( $config[ $key ][ $module ] ?? [], $settings );
+					// Check config settings syntax is valid.
```
roborourke: I think this comment would make sense just before the `if` statement below it. Here we're just assigning the value to a variable. Minor thing really but makes it a little clearer.

inc/namespace.php
```diff
@@ -80,7 +80,23 @@ function merge_config_settings( array $config, array $overrides ) : array {
...
+					$config_setting = $config[ $key ][ $module ];
```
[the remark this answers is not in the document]
roborourke: In keeping with the code comments and the name of the variable we're merging this with lets keep it consistent and name the variable `$config_settings`.

roborourke:
> Changing values in the composer.json and then looking at php_errors.log felt a bit temperamental and still displaying the error when I "think" it should have worked.

I believe this is because you removed the null coalesce part when getting the value from the config: `?? []` - this ensures if there's nothing set or the value is explicitly `null` that we get an array back.

>  let me know how I can correctly test this

On the original issue if you look at Dzikri's example you should see that _without_ your changes a config like this throws a warning about the `array_merge` arguments.

```json
{
  "extra": {
    "altis": {
      "modules": {
        "workflow": false
      }
    }
  }
}
```

Once your changes are applied the warning should be gone and no other new warnings thrown.

inc/namespace.php
dashaluna: Good point, tbh I looked at how other code outputs the `trigger_error` and copied the approach.
Will keep in mind the rule specifics.

inc/namespace.php
dashaluna: Ah, I did think about that :) changing...

dashaluna:
@roborourke thanks! Fixed.

Re testing, yes I used the same snipped. I got confused when I tried wrong values, to see our custom warning displayed, but instead I can still see `PHP Warning:  array_merge(): Argument #2 is not an array` one instead of the custom one.

This is for example, using the following config, where value is a `string` (and not `array` or `bool`):

```
{
  "extra": {
    "altis": {
      "modules": {
        "workflow": 'false'
      }
    }
  }
}
```

inc/namespace.php
```diff
@@ -80,7 +80,22 @@ function merge_config_settings( array $config, array $overrides ) : array {
 			// Merge module settings together.
 			case 'modules':
 				foreach ( $value as $module => $settings ) {
-					$config[ $key ][ $module ] = array_merge( $config[ $key ][ $module ] ?? [], $settings );
+					$config_settings = $config[ $key ][ $module ] ?? [];
+
+					// Check config settings syntax is valid.
+					if ( ! is_array( $config_settings ) && ! is_bool( $config_settings ) ) {
```
roborourke: 🤦‍♂ I should've spotted this - the clue is in the error message, it's actually the 2nd argument in the `array_merge` that's the problem!

So change all of these to be checks against the `$settings` variable.

inc/namespace.php
```diff
@@ -80,7 +80,22 @@ function merge_config_settings( array $config, array $overrides ) : array {
 			// Merge module settings together.
 			case 'modules':
 				foreach ( $value as $module => $settings ) {
-					$config[ $key ][ $module ] = array_merge( $config[ $key ][ $module ] ?? [], $settings );
+					$config_settings = $config[ $key ][ $module ] ?? [];
```
roborourke: Let's move this back to the `array_merge()` call, we don't need to use this variable anymore.

inc/namespace.php
dashaluna: Oops, I didn't notice that too! 🤦‍♂

inc/namespace.php
```diff
@@ -80,6 +80,20 @@ function merge_config_settings( array $config, array $overrides ) : array {
 			// Merge module settings together.
 			case 'modules':
 				foreach ( $value as $module => $settings ) {
+
+					// Check settings syntax is valid.
+					if ( ! is_array( $settings ) && ! is_bool( $settings ) ) {
```
kucrut: @roborourke @dashaluna Shouldn't this be a `||` instead of `&&`?

roborourke: No, if `$settings` was a boolean then the `! is_array()` check would return true and the if statement would be executed. It has to be not an array _and_ not a boolean to trigger the warning.

PR #68, opened 2020-02-05
Add get_environment_region() function

### LINKED ISSUE #65: Add get_environment_region() function
We have functions to get environment name, type and architecture. We should also have one to get the region.

Acceptance criteria:

- [x] Add a function called `get_environment_region()` that returns `HM_ENV_REGION` if set, otherwise `null`
- [x] Function has a return type hint of `: ?string`
- [x] Function is documented in functions section of `docs/readme.md`

dashaluna:
Fixes #65

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -20,7 +20,7 @@ Returns one of `local`, `development`, `staging` or `production`. This is read f
 
 **`get_environment_region() : ?string`**
 
-Returns the current AWS application's region based on `HM_ENV_REGION` constant. On local setups or if the region is not set this will return `null`. The region can be `us-east-1`, `ap-northeast-1`, `eu-central-1` etc.
+Returns the current AWS application's region based on the `HM_ENV_REGION` constant. On local setups or if the region is not set this will return `null`. The region can be `us-east-1`, `ap-northeast-1`, `eu-central-1` etc.
 
 **`get_environment_architecture() : string`**
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -20,7 +20,7 @@ Returns one of `local`, `development`, `staging` or `production`. This is read f
 
 **`get_environment_region() : ?string`**
 
-Returns the current hosting stack region based on `HM_ENV_REGION` constant. On local setups and if region is not set this will return `null`.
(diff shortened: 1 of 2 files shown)


inc/namespace.php
```diff
@@ -193,6 +193,19 @@ function get_environment_type() : string {
 	return 'local';
 }
 
+/**
+ * Get the region of the current environment.
+ *
+ * @return string|null Region if set,
+ *                     null otherwise.
```
roborourke: No need for the line wrap here

inc/namespace.php
```diff
@@ -193,6 +193,19 @@ function get_environment_type() : string {
 	return 'local';
 }
 
+/**
+ * Get the region of the current environment.
```
roborourke: Thinking about it we could make this a bit clearer that it's the AWS region:
```suggestion
 * Get the AWS region of the current environment.
```

Also add some examples like in the doc block for the function above eg. `us-east-1`, `ap-northeast-1`, `eu-central-1` etc

docs/README.md
```diff
@@ -18,6 +18,10 @@ Returns the current hosting stack name based on the `HM_ENV` constant. On local
 
 Returns one of `local`, `development`, `staging` or `production`. This is read from the `HM_ENV_TYPE` constant except on local setups.
 
+**`get_environment_region() : ?string`**
+
+Returns the current hosting stack region based on `HM_ENV_REGION` constant. On local setups and if region is not set this will return `null`.
```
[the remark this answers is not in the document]
dashaluna: Adding `Returns the current AWS application's region...` does that sound ok? Even for better clarity on what application.

docs/README.md
```diff
@@ -18,6 +18,10 @@ Returns the current hosting stack name based on the `HM_ENV` constant. On local
 
 Returns one of `local`, `development`, `staging` or `production`. This is read from the `HM_ENV_TYPE` constant except on local setups.
 
+**`get_environment_region() : ?string`**
+
+Returns the current AWS application's region based on `HM_ENV_REGION` constant. On local setups or if the region is not set this will return `null`. The region can be `us-east-1`, `ap-northeast-1`, `eu-central-1` etc.
```
roborourke:
```suggestion
Returns the current AWS application's region based on the `HM_ENV_REGION` constant. On local setups or if the region is not set this will return `null`. The region can be `us-east-1`, `ap-northeast-1`, `eu-central-1` etc.
```

roborourke:
🚀

PR #75, opened 2020-02-10
Fix config path for aws credentials

joehoyle:
Turns out we missed the `modules` sub-key in the `config` object. Fixing. I'm pretty sure this was never used anywhere, but can be useful if you have keys and want to use the AWS SDK locally.


roborourke:
We didn't support config for credentials in v1 so don't think we need to backport that.

PR #78, opened 2020-02-13
Add ElasticPress debug bar (to exclusion for WP installer)

dashaluna:
Part of https://github.com/humanmade/altis-enhanced-search/issues/55


inc/composer/class-installer.php
```diff
@@ -25,6 +25,7 @@ public function getInstallPath( PackageInterface $package, $framework_type = ''
 		 */
 		$excluded_plugins = [
 			'10up/elasticpress',
+			'10up/debug-bar-elasticpress',
```
roborourke: Minor thing, these are in alphabetical order, so d before e

PR #81, opened 2020-03-17
Throw an exception for invalid config

### LINKED ISSUE #21: Throw Exception on invalid config
I've noticed if you have an error in your config, we don't `die`, and if that goes unnoticed, it's very confusing! I think we should a big bad red error for this!

-------------

Acceptance criteria:

- [ ] Change `E_USER_WARNING` to `E_USER_ERROR` in `get_json_file_contents_as_array()` on line 123
- [ ] Add text to the error message before `json_last_error_msg()` saying "`composer.json` could not be parsed\\n"
- [ ] Validate module names under the modules property against registered modules and allow unknown modules if they have an `entrypoint` property, trigger an error if invalid module found
- [ ] Check module names against top level config properties eg. `extra.altis.<module>` and trigger an error if a module name is found with instructions to move it under `extra.altis.modules.<module>`
- [ ] On local environments instead of just triggering an error throw an exception with the same error message

https://github.com/humanmade/altis-core/blob/f0d4e1a1230a149d6c84c99756657cbf7bdcdd71/inc/namespace.php#L119-L124

Use `Altis\Module::get_all()` to get all registered modules.

Examples of invalid configs:

Unregistered module with no "entrypoint" property:
```json
{
  "extra": {
    "altis": {
      "modules": {
        "enhanced-search": { "enabled": false }
      }
    }
  }
}
```

Registered module config not under "modules" property:
```json
{
  "extra": {
    "altis": {
      "search": { "enabled": false }
    }
  }
}
```

Custom module at top level with "entrypoint" property:
```json
{
  "extra": {
    "altis": {
      "custom-module": { "entrypoint": "path/to/file.php" }
    }
  }
}
```

ivdimova:
Opening PR to be able to comment here. 

Fixes #21

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -80,7 +80,7 @@ function validate_config_settings( array $config ) {
 	$error_type = $environment === 'local' ? E_USER_ERROR : E_USER_WARNING;
 	// Get configs to validate.
 	$configs = [
-		'default' => $config
+		'default' => $config,
 	];
 	$registered_modules = Module::get_all();
 
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -58,7 +58,7 @@ function get_merged_config() : array {
 	$composer_file = ROOT_DIR . '/composer.json';
 
 	$composer_json = get_json_file_contents_as_array( $composer_file );
-
+	validate_config_settings( $composer_json );
 	$config = merge_config_settings( $default_config, $composer_json['extra']['altis'] ?? [] );
 
 	// Look for environment specific settings in the config and merge it in.
@@ -66,6 +66,19 @@ function get_merged_config() : array {
 	$config = merge_config_settings( $config, $config['environments'][ $environment ] ?? [] );
 
 	return $config;
+}
+
+function validate_config_settings( $composer_json ) {
+	$registered_modules = Module::get_all();
+
+	$modules[] = $composer_json['extra']['altis']['modules'];
+
+	foreach ( $modules as $module_name => $module ) {
+		if ( ! array_key_exists( $module_name, $registered_modules ) && ! array_key_exists( 'entrypoint', $module ) ) {
+			// phpcs:ignore
+			trigger_error( 'Custom modules should have entrypoint property! Your module ' . $module_name . ' is missing the entrypoint property!', E_USER_WARNING );
+		}
+	}
 }
 
 /**
@@ -118,7 +131,6 @@ function merge_config_settings( array $config, array $overrides ) : array {
  * @return array Decoded data in array form, empty array if JSON data could not read.
  */
 function get_json_file_contents_as_array( $file ) : array {
-
 	if ( ! strpos( $file, '.json' ) ) {
 		// phpcs:ignore
 		trigger_error( $file . ' is not a JSON file.', E_USER_ERROR );
@@ -138,23 +150,7 @@ function get_json_file_contents_as_array( $file ) : array {
 		trigger_error( "composer.json could not be parsed\n" . json_last_error_msg(), E_USER_WARNING );
 		return [];
 	}
-	$registered_modules = Module::get_all();
-	foreach ( $contents['extra']['altis'] as $config_prop ) {
-		if ( array_key_exists( $config_prop[0], $registered_modules ) ) {
-			// phpcs:ignore
-			trigger_error( 'Modules should be listed under modules property!', E_USER_WARNING );
-			return [];
-		}
-	}
-	foreach ( $contents['extra']['altis']['modules'] as $module ) {
-		if ( ! in_array( $module, $registered_modules ) ) {
-			if ( ! array_key_exists( 'entrypoint', $module ) ) {
-				// phpcs:ignore
-				trigger_error( 'Custom modules should have entrypoint property!', E_USER_WARNING );
-				return [];
-			}
-		}
-	}
+
 	return $contents;
 }
 
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -94,35 +94,33 @@ function validate_config_settings( array $config ) {
 		$configs[ $environment ] = merge_config_settings( $config, $config['environments'][ $environment ] ?? [] );
 	}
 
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -56,17 +56,34 @@ function get_merged_config() : array {
...
 	$config = merge_config_settings( $default_config, $composer_json['extra']['altis'] ?? [] );
 
 	// Look for environment specific settings in the config and merge it in.
-	$environment = get_environment_type();
 	$config = merge_config_settings( $config, $config['environments'][ $environment ] ?? [] );
 
 	return $config;
 }
 
+function validate_config_settings( $composer_json, $environment ) {
```
roborourke: Try and get in the habit of using type hints with your functions, helps to catch errors and makes development clearer and easier.

```suggestion
function validate_config_settings( array $composer_json, string $environment ) {
```

inc/namespace.php
```diff
@@ -56,17 +56,34 @@ function get_merged_config() : array {
 
 	// Get custom config overrides.
 	$composer_file = ROOT_DIR . '/composer.json';
-	$composer_json = get_json_file_contents_as_array( $composer_file );
 
+	$composer_json = get_json_file_contents_as_array( $composer_file );
+	$environment = get_environment_type();
+	validate_config_settings( $composer_json, $environment );
```
roborourke: What was the thinking behind passing in the full `$composer_json` data? Seems we only need the data in `$config` to me

I'm thinking we should do the validation in 2 places:

1. Validate the default config _before_ the environment overrides are merged in (as you are doing)
2. Validate the config again after merging environment type overrides

This way we can direct people to where the problem is a little better for example "There is a problem in the default config" or "There is a problem in the staging config".

I think we can do the per environment validation within the `validate_config_settings()` function itself eg:

```php
function validate_config_settings( array $config ) {
     $environment = get_environment_type();

     // Get configs to validate.
     $configs = [ 
         'default' => $config
     ];

     // Validate all environment types locally to help avoid pushing unforeseen problems to production.
     if ( $environment === 'local' ) {
          $configs['local'] = merge_config_settings( $config, $config['environments']['local'] ?? [] );
          $configs['development'] = merge_config_settings( $config, $config['environments']['development'] ?? [] );
          $configs['staging'] = merge_config_settings( $config, $config['environments']['staging'] ?? [] );
          $configs['production'] = merge_config_settings( $config, $config['environments']['production'] ?? [] );
     } else {
          $configs[ $environment ] = merge_config_settings( $config, $config['environments'][ $environment ] ?? [] );
     }

     foreach ( $configs as $type => $conf ) {
         // Do config validation here, adding the config type name to the error messages.
     }
}
```

ivdimova: So I finally was able to sit down and look at this and I have a few questions. First of all - I've done what @roborourke suggested, however I am not sure it couldn’t be easier. Merging all the environment configs and validating after this seems a bit overkill - the code will actually only merge the modules from each config so we can check these without the need of merging and then looping through them. That said, I don't see a better way to determine if there is module under the environment config that is not under `modules` property. I am open to suggestions for these, I might be missing something.

roborourke: I'm only suggesting doing that on local so you can see any issues. The reason is that the config overrides for a given environment might not have issues once merged with the base config so we should only validate after merging. I don't think it's overkill in that case if it prevents errors getting to production stacks.

ivdimova: @roborourke  I am not saying that it's overkill to do it, just wondering if it's worth doing it this way considering that it seems only modules get merged. In any case, if you think this is best, I am ok with it.

roborourke: Right sorry, so that's not entirely the case - if you have a top-level config item it will be overridden by any matching top level items in an environment specific config. Modules are a special case as they're merged one level deeper. We'll improve this over time but currently pretty much all config that needs to be merged and checked is in the modules section

inc/namespace.php
```diff
@@ -56,17 +56,73 @@ function get_merged_config() : array {
...
+	} else {
+		$configs[ $environment ] = merge_config_settings( $config, $config['environments'][ $environment ] ?? [] );
+	}
+
+	foreach ( $configs as $type => $conf ) {
+		foreach ( $conf['modules'] as $module_name => $module ) {
+			if ( ! array_key_exists( $module_name, $registered_modules ) && ! array_key_exists( 'entrypoint', $module ) ) {
+				trigger_error(
+					sprintf(
+						'Custom modules should have entrypoint property! Your module %1$s is missing the entrypoint property at %2$s environment!',
```
roborourke:
```suggestion
						'Custom modules should have an entrypoint property. Your module %1$s is missing the entrypoint property in your %2$s settings.' . "\n" .
						'Custom module documentation: https://www.altis-dxp.com/resources/docs/getting-started/custom-modules/',
```

inc/namespace.php
```diff
@@ -56,17 +56,73 @@ function get_merged_config() : array {
...
+						$module_name,
+						$type
+					),
+					$error_type
+				);
+			}
+		}
+
+		foreach ( $conf as $key => $value ) {
+			if ( $key !== 'modules' && $key !== 'environments' ) {
```
roborourke: So here we want to check if the top level `$key` is a registered module name that we know rather than just checking if it's not `modules` or `environments`. We want to catch errors where a known module name is being used incorrectly at the top level rather than any arbitrary properties - we do want to allow those. Does that make sense?

So it'd be something like:

```suggestion
			if ( array_key_exists( $key, $registered_modules ) ) {
```

inc/namespace.php
```diff
@@ -56,17 +56,73 @@ function get_merged_config() : array {
...
+				trigger_error(
+					sprintf(
+						'All modules should be under modules property. Your %1$s module is at top level property at %2$s environment!',
```
roborourke:
```suggestion
						'All module settings should be under the modules property. Settings for the module %1$s were found at the top level in the %2$s settings.',
```

roborourke (approved):

PR #82, opened 2020-06-04
Fix linter errors

roborourke:
This fixes the current HM Linter errors for the module.

### DIFF
--- a/inc/about/namespace.php
+++ b/inc/about/namespace.php
@@ -1,9 +1,14 @@
 <?php
+/**
+ * Altis About Page.
+ *
+ * @package altis
+ */
 
 namespace Altis\About;
 
+use Altis;
 use Altis\Module;
-use function Altis\get_composer_data;
 use WP_Admin_Bar;
 
 const PAGE_SLUG = 'altis-about';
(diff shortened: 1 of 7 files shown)


roborourke:
@rmccue this one is good to go now.

inc/about/namespace.php
```diff
@@ -204,7 +209,7 @@ function render_credits_page() {
 			<p>Altis includes other libraries as part of WordPress.</p>
 			<?php
 			array_walk( $libraries['data'], '_wp_credits_build_object_link' );
-			echo '<p class="wp-credits-list">' . wp_sprintf( '%l.', $libraries['data'] ) . "</p>\n\n";
+			echo '<p class="wp-credits-list">' . sprintf( '%s.', wp_kses( implode( ', ', $libraries['data'] ), [ 'a' => [ 'href' => [] ] ] ) ) . "</p>\n\n";
```
rmccue, line 212: Could use `whitelist_html` here instead now that's in Altis? Also, no need for `sprintf( '%s'`, can move the `.` back into the string at the end.

roborourke: Oh yeah duh, the problem is our coding standards aren't aware of `whitelist_html()` as an escaping function.

rmccue: Ah! Added that in HM-Minimum I think

PR #83, opened 2020-07-06
Config validation

### LINKED ISSUE #21: Throw Exception on invalid config

pdewouters:
fixes #21 

The errors get logged to the PHP error log, but may still cause a white screen because of the error level

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -169,7 +169,7 @@ function get_json_file_contents_as_array( $file ) : array {
  * @param string $msg
  * @param int $level
  */
-function log_error( string $msg, int $level = E_USER_ERROR ) {
+function log_error( string $msg, int $level = E_USER_WARNING ) {
 	if ( get_environment_type() === 'local' ) {
 		throw new RuntimeException( $msg );
 	}
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -120,7 +120,7 @@ function merge_config_settings( array $config, array $overrides ) : array {
 function get_json_file_contents_as_array( $file ) : array {
 	if ( ! strpos( $file, '.json' ) ) {
 		// phpcs:ignore
-		trigger_error( $file . ' is not a JSON file.', E_USER_WARNING );
+		trigger_error( $file . ' is not a JSON file.', E_USER_ERROR  );
 		return [];
 	}
 
@@ -135,7 +135,7 @@ function get_json_file_contents_as_array( $file ) : array {
 
 	if ( json_last_error() !== JSON_ERROR_NONE ) {
 		// phpcs:ignore
-		trigger_error( json_last_error_msg(), E_USER_WARNING );
+		trigger_error( 'composer.json could not be parsed\n' . json_last_error_msg(), E_USER_WARNING );
 		return [];
 	}
 
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -8,6 +8,7 @@
 namespace Altis;
 
 use Aws\Sdk;
+use RuntimeException;
 
 /**
  * Bootstrap any core functions as necessary.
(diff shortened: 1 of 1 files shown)


roborourke:
If they're not logged I'm not sure we can say it's solved - I'm not sure I follow what you mean in your comment to be honest.

There was already an open PR for this here #81 - I'll check over what you have here but we may want to take elements of what you've done and apply it there to complete the work.

inc/namespace.php
```diff
@@ -135,13 +135,48 @@ function get_json_file_contents_as_array( $file ) : array {
 
 	if ( json_last_error() !== JSON_ERROR_NONE ) {
 		// phpcs:ignore
-		trigger_error( json_last_error_msg(), E_USER_WARNING );
+		trigger_error( 'composer.json could not be parsed\n' . json_last_error_msg(), E_USER_WARNING );
```
roborourke, line 138: Worth noting that the newline character `\n` will not be parsed as a newline unless it's inside double quotes

inc/namespace.php
```diff
@@ -135,13 +135,48 @@ function get_json_file_contents_as_array( $file ) : array {
...
 	return $contents;
 }
 
+/**
+ * Throws exception or triggers error depending on environment.
+ *
+ * @param string $msg
+ * @param int $level
+ */
+function log_error( string $msg, int $level = E_USER_ERROR ) {
```
roborourke, line 172: `E_USER_ERROR` will cause PHP to stop executing, hence why we use `E_USER_WARNING` generally

pdewouters:
@roborourke sorry, I didn't see that existing PR. I don' think the issue was in the correct column

The errors are actually logged and visible via `composer server logs php` 
`php_1              | [06-Jul-2020 14:52:41 UTC] PHP Fatal error:  Uncaught RuntimeException: Please specify an entry point setting for your custom module. in /usr/src/app/packages/core/inc/namespace.php:174`

inc/namespace.php
rmccue: We should avoid newlines in error messages anyway (as they can make parsing harder later), suggest:
```suggestion
		trigger_error( sprintf( 'composer.json could not be parsed: %s', json_last_error_msg() ), E_USER_WARNING );
```

jazzsequence:
Updated this PR to fix the couple things that came out of code review.

roborourke:
Ah sorry @jazzsequence - I was looking at #81 as the solution here, it goes into a bit more depth and uses the merged config to determine validity. I should have closed this one out.

jazzsequence:
@roborourke no worries, I saw it sitting in the code review column so I thought I'd take a look. I'll peek over there. 👀

roborourke:
Closing in favour of #81

PR #86, opened 2020-07-29
Add post cloner plugin to allowlist

pdewouters:
This will allow for https://github.com/humanmade/altis-cms/pull/248 - including the post cloner plugin to the CMS module

Doh, old habits die hard, forgot to use "allowlist" in the commit message


roborourke:
Don't worry about it. I think we can rename the variable in that file too, probably what prompted it :)

PR #92, opened 2020-08-25
[88] add altis consent

jazzsequence:
Adds the consent module to `altis-core`.

Loads the consent API from `altis/consent-api` and the consent module from `altis/consent`


jazzsequence:
@roborourke 
I'm thinking about ways to use the Altis config file to adjust/change the settings. A couple things I have questions about:

There are some settings (notably the banner message) that I'm defaulting to an empty string because we don't have access to escaping or i18n functions when the config is loaded. That's fine, but it means if you pass something into that setting from the config, it can't be internationalized. So, I'm sort of using that as a default setting, so changes are saved in the admin.

That sort of defeats the purpose (it seems) of configuring stuff in the composer file, though, so I'm wondering if there should be a different way of handling it or perhaps fewer configurable options?

Another thought: I'd like to be able to _disable_ settings from the config. I'm thinking if you put something like:

```json
"consent": [
    "banner_options": false
]
```
it would remove that setting in the admin. However, there are some settings that are saved as a boolean, so that breaks that theory. I'm wondering if something like this could work without being too confusing:

```json
"consent": [
    "banner_options": [
        "value": "none",
        "display": false
    ]
]
```

My thought is you could set `"banner_options": "none"` or whatever on one line, or pass in an array with `"display"` and if `display` is set to false, it hides the option in the admin.

Am I rabbit-holing too much? Does this make sense?

roborourke:
@jazzsequence sorry for the delay getting back to you here.

So first thing, any actual text to be shown to users isn't a suitable candidate for composer.json config as you rightly state. We're looking more at config options that affect the behaviour of the feature.

With regards to setting options at a higher level vs lower level we already have precedent for this approach, for example with the publication checklist or notifications in the Workflow module:

https://github.com/humanmade/altis-workflow/blob/master/inc/namespace.php#L34-L49

You're essentially checking for truthy values, then checking the value type to see if it's an array, at which point you can set default values if it's not an array, and then check for specific keys within that array to continue configuring the plugin. You can use those defaults or provided config options to build the logic for which filters to add/remove etc.

In the Analytics module we use the presence of a configured tag ID to decide whether or not to remove the action hooks that add the admin UI:

https://github.com/humanmade/altis-analytics/blob/master/inc/google_tag_manager/namespace.php#L15-L28

So, I'd say you're rabbit holing a little bit with regards to handling the `banner_options` value specifically however you're definitely thinking in the right way about how configuration can and should affect the behaviour of the plugin, e.g. configured values should not be editable in the admin.

inc/consent/namespace.php
```diff
@@ -0,0 +1,57 @@
...
+
+	// Bail if we've turned consent off explicitly.
+	if ( $config === false ) {
+		return;
+	}
+
+	// Bail if options have been set.
+	if ( $options ) {
+		return;
+	}
```
roborourke, line 31: I think this might be confusing, in my mind even if I'd set some options if I was to then configure it I'd expect the configuration to take precedence. It'd be hard to debug what was happening and why otherwise.

inc/consent/namespace.php
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Save the defaults to the database if nothing has been set yet.
+ */
+function set_consent_defaults() {
+	$config  = Altis\get_config()['modules']['core']['consent'];
+	$options = get_option( 'cookie_consent_options' );
+
+	// Bail if we've turned consent off explicitly.
+	if ( $config === false ) {
```
roborourke, line 24: Should this be the same check as in `load_plugins()`? Say in some weird circumstance that `$config` was `null` this code would still execute but the plugins themselves wouldn't be loaded.
```suggestion
	if ( empty( $config ) ) {
```

load.php
```diff
@@ -29,6 +29,14 @@
 		'Core',
 		[
 			'enabled' => true,
+			'consent' => [
+				'display_banner'      => false,
+				'privacy_policy_page' => '',
+				'cookie_policy_page'  => '',
+				'cookie_expiration'   => 30,
+				'banner_options'      => 'none',
+				'banner_text'         => '',
```
roborourke, line 38: It's been a little inconsistent but the bulk of our config options use hyphens rather than underscores. I think that was never standardised in the config RFC so I should add an RFC for standardising these things.

roborourke (changes requested):
Couple of immediate changes we'll need to make here. One of the bigger hurdles is getting the plugins finalised, versioned and on packagist. We might need to either fork the consent API plugin or see if they would be willing to publish it very soon.

composer.json
```diff
@@ -9,16 +9,29 @@
...
+			"url": "git@github.com:rlankhorst/wp-consent-level-api.git"
+		},
+		"altis/consent": {
+			"type": "vcs",
+			"url": "git@github.com:humanmade/altis-consent.git"
+		}
+	},
 	"require": {
 		"php": ">=7.1",
+		"altis/consent": "*",
```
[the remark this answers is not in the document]
jazzsequence: This was a placeholder for testing since the plugin itself needs an updated release. I will update that as soon as I tag it. The same applies to the vcs repository for `altis/consent` above. ☝️

composer.json
```diff
@@ -9,16 +9,29 @@
...
 		"aws/aws-sdk-php": "^3.150.0",
 		"composer-plugin-api": "^1.0",
-		"composer/installers": "^1.4"
+		"composer/installers": "^1.4",
+		"rlankhorst/wp-consent-level-api": "dev-master as 1.0.0"
```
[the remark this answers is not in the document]
jazzsequence: I've asked about tagging releases and that ticket hasn't gotten any feedback (but other tickets and PRs I've submitted have), so we may want to just fork it.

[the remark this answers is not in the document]
jazzsequence: I've opened an issue in their repo in the meantime https://github.com/rlankhorst/wp-consent-level-api/issues/59

composer.json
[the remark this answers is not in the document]
jazzsequence: When the `altis-consent` plugin is good to go in https://github.com/humanmade/altis-consent/pull/18, I'll make a release and update this.

inc/consent/namespace.php
jazzsequence: Yep, the way I wrote this originally was setting things in the config was a default value, but it's been refactored now.

jazzsequence:
Alright, this should be good with the one caveat that the version for the `altis/consent` package is dependent upon finalizing the remaining issues (notably https://github.com/humanmade/altis-consent/pull/18) in that repo and then making a new release.

inc/composer/class-installer.php
```diff
@@ -71,6 +72,7 @@ public function getInstallPath( PackageInterface $package, $framework_type = ''
 			'humanmade/wp-seo',
 			'humanmade/wp-simple-saml',
 			'johnbillion/query-monitor',
+			'rlankhorst/wp-consent-level-api',
```
roborourke, line 75: Seems this is now `altis/consent-api`?

jazzsequence: 🤦‍♂️ d'oh, forgot that one 👍

jazzsequence: @roborourke fixed now 🙄

composer.json
```diff
@@ -11,14 +11,17 @@
 	],
 	"require": {
 		"php": ">=7.1",
+		"altis/consent-api": "^0.9",
```
roborourke, line 14: Just occurred to me but this should be a dependency of the consent plugin rather than a direct dependency here

jazzsequence: I wasn't sure how that would work since it needs to be in the main `vendor` directory rather than the consent plugin's `vendor` directory. I assume the [installation handling](https://github.com/humanmade/altis-core/pull/92/files#diff-093e57f43775d5bb8a825af36fbd068bR46) that's done in `class-installer` would handle that even if it's further down the tree?

roborourke: Yep correct, that list of packages is the complete list once Composer has figured out the entire tree.

jazzsequence: Cool cool. I'll update that. 👍

jazzsequence: https://github.com/humanmade/altis-consent/pull/21

jazzsequence:
This should now block first party analytics and gtm if not consented to.

inc/consent/namespace.php
```diff
@@ -0,0 +1,115 @@
...
+		if ( is_numeric( $config['cookie-expiration'] ) ) {
+			$options['cookie_expiration'] = (int) $config['cookie_expiration'];
+			unset( $fields['cookie_expiration'] );
+		}
+	}
+
+	// If any options were set in the config, update those values in the database and remove the options on the settings page.
+	if ( ! empty( $options ) ) {
+		update_option( 'cookie_consent_option', $options );
+		add_filter( 'altis.consent.consent_settings_fields', $fields );
```
roborourke, line 68: How does this filter work? The 2nd param should be a callback but `$fields` is a keyed array.

inc/consent/namespace.php
```diff
@@ -0,0 +1,115 @@
...
+		// Make sure the value set in the config is numeric.
+		if ( is_numeric( $config['cookie-expiration'] ) ) {
+			$options['cookie_expiration'] = (int) $config['cookie_expiration'];
+			unset( $fields['cookie_expiration'] );
+		}
+	}
+
+	// If any options were set in the config, update those values in the database and remove the options on the settings page.
+	if ( ! empty( $options ) ) {
+		update_option( 'cookie_consent_option', $options );
```
roborourke, line 67: Given that this can run on `plugins_loaded` it will cause issues - it will run on every load meaning every request makes database updates. This can also cause issues during installation because this code can run before the database tables have been created.

The right solution for setting configured values is to filter the option value to return the configured one.

inc/consent/namespace.php
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/**
+ * Set the analytics script to no-op if no consent to statistics has been granted.
+ *
+ * @param bool $noop No-op value to prevent analytics events from being sent to Pinpoint. Defaults to false.
+ *
+ * @return bool      The updated noop value.
+ */
+function set_analytics_noop( bool $noop ) : bool {
```
roborourke, line 92: Just a note we might still have to come back to this and just unregister the script altogether. I _think_ it'll be ok as long other dependent code has the analytics script set as a dependency. This is the issue I was referring to on our chat the other day https://github.com/humanmade/platform-dev/issues/550

We'll see though. A front end solution would be to set the value of `Altis.Analytics.Noop` directly on the client side.

inc/consent/namespace.php
```diff
@@ -0,0 +1,115 @@
+<?php
+
+namespace Altis\Core\Consent;
```
roborourke, line 3: We just use the `Altis` namespace for core stuff, `Altis\Core` isn't used anywhere.
```suggestion
namespace Altis\Consent;
```

inc/namespace.php
```diff
@@ -14,6 +14,7 @@
  */
 function bootstrap() {
 	About\bootstrap();
+	Core\Consent\bootstrap();
```
roborourke, line 17: We don't have a specific `Core` namespace, just the root namespace `Altis`
```suggestion
	Consent\bootstrap();
```

roborourke (changes requested):
Main thing to update is the implementation of the default config values. Will dig into the analytics no-op issue a bit further and get back to you.

inc/consent/namespace.php
```diff
@@ -0,0 +1,115 @@
...
+ */
+function bootstrap() {
+	add_action( 'plugins_loaded', __NAMESPACE__ . '\\set_consent_options' );
+	add_action( 'plugins_loaded', __NAMESPACE__ . '\\load_plugins', 1 );
+
+	add_filter( 'altis.analytics.noop', __NAMESPACE__ . '\\set_analytics_noop' );
+	add_filter( 'altis.analytics.data', __NAMESPACE__ . '\\set_analytics_data', 1 );
+
+	// If statistics hasn't been consented to, don't load the GTM output.
+	if ( ! wp_get_consent( 'statistics' ) ) {
```
roborourke, line 19: `wp_get_consent()` doesn't appear to be a function 😛 
```suggestion
	if ( ! wp_get_consent_type( 'statistics' ) ) {
```

roborourke (changes requested):

inc/consent/namespace.php
```diff
@@ -0,0 +1,115 @@
...
+			unset( $fields['display_banner'] );
+		}
+	}
+
+	// Check if banner-options was configured.
+	if ( ! empty( $config['banner_options'] ) ) {
+		// Make sure any option set in the config is a valid option.
+		if ( in_array( $config['banner-options'], Settings\get_cookie_banner_options(), true ) ) {
+			$options['banner_options'] = $config['banner-options'];
+			unset( $fields['banner_options'] );
```
rmccue, line 52: These are checking inconsistently named options, using `$config['banner_options']` at the start but `$config['banner-options']` later

load.php
```diff
@@ -29,6 +29,11 @@
 		'Core',
 		[
 			'enabled' => true,
+			'consent' => [
+				'display-banner'      => '',
+				'cookie-expiration'   => '',
+				'banner-options'      => '',
```
rmccue, line 35: `null` seems like a better choice for these defaults I think?

Also, shouldn't be aligned, as it causes code churn in the future.

inc/consent/namespace.php
rmccue, line 68: This causes an error as it's not a function; should be `'__return_empty_array'` assuming you're removing the options from the UI here.

inc/consent/namespace.php
rmccue, line 67: As @roborourke noted, this should be using a filter instead. Also, this appears to be the wrong option name, as it's actually `cookie_consent_options` - this should really be prefixed though.

```suggestion
		add_filter( 'pre_option_cookie_consent_option', function () use ( $options ) {
			return $options;
		} );
```

rmccue (dismissed):

inc/consent/namespace.php
```diff
@@ -0,0 +1,115 @@
...
+		if ( is_bool( $config['display-banner'] ) ) {
+			$options['display_banner'] = $config['display-banner'];
+			unset( $fields['display_banner'] );
+		}
+	}
+
+	// Check if banner-options was configured.
+	if ( ! empty( $config['banner_options'] ) ) {
+		// Make sure any option set in the config is a valid option.
+		if ( in_array( $config['banner-options'], Settings\get_cookie_banner_options(), true ) ) {
```
rmccue, line 50: This compares the string in `banner-options` with the deeper `<select>`-style options array returned from this function, so doesn't match.
```suggestion
		if ( in_array( $config['banner-options'], wp_list_pluck( Settings\get_cookie_banner_options(), 'value' ), true ) ) {
```

inc/consent/namespace.php
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Update the Endpoint/User data if we don't have consent for statistics.
+ * Overrides with an empty array if we don't have consent.
+ *
+ * @param array $data An array of analytics data to send to Pinpoint.
+ *
+ * @return array      The filtered analytics variable data.
+ */
+function set_analytics_data( array $data ) : array {
```
rmccue, line 109: `$data` is a `stdClass` here, not array.

inc/consent/namespace.php
```diff
@@ -0,0 +1,115 @@
...
+	if ( ! wp_has_consent( 'statistics' ) ) {
+		$data['Endpoint']['User'] = [];
```
rmccue, line 111: `$data['Endpoint']` is an object here, not an array.
```suggestion
		$data['Endpoint']->User = [];
```

inc/consent/namespace.php
jazzsequence: I added the `Core` to distinguish it from `Altis\Consent` because that would introduce a function name clash with `Altis\Consent\bootstrap`

jazzsequence: I added `Altis\Core\` because there is already an [`Altis\Consent\bootstrap()`](https://github.com/humanmade/altis-consent/blob/master/inc/namespace.php#L18) which would clash with [`Altis\Core\Consent\bootstrap()`](https://github.com/humanmade/altis-core/blob/12b0c1aedd2d1492af48a399955cd9decc3f239c/inc/consent/namespace.php#L11) if left in the `Altis\Consent` namespace. 

I wasn't sure what the best way to handle this clash was other than creating a `Altis\Core` namespace. I considered renaming the function `init()`, but I was afraid of other potential clashes with `Altis\Consent` if I left the `Core` module using the same namespace.

inc/namespace.php
jazzsequence: See above https://github.com/humanmade/altis-core/pull/92#discussion_r488944265

inc/consent/namespace.php
roborourke: Ah I see, this seems like a reasonable workaround then 👍

inc/composer/class-installer.php
```diff
@@ -42,6 +42,7 @@ public function getInstallPath( PackageInterface $package, $framework_type = ''
 			'10up/elasticpress',
 			'altis/aws-analytics',
 			'altis/browser-security',
+			'altis/consent',
```
roborourke, line 45: Need to add the consent API JS plugin here too, just checked and it's also of the type `wordpress-plugin` but expected to be loaded from the vendor folder.

```suggestion
			'altis/consent',
			'altis/consent-api',
```

composer.json
```diff
@@ -11,14 +11,16 @@
 	],
 	"require": {
 		"php": ">=7.1",
+		"altis/consent": "^1.0.0",
```
roborourke, line 14: For our own plugins at least we should keep the version constraint tighter so that only patch releases are allowed.
```suggestion
		"altis/consent": "~1.0.0",
```

roborourke (approved):
Thanks @jazzsequence, approved pending the 2 suggestions I added 👍

inc/composer/class-installer.php
jazzsequence: Yep. I knew this, but for some reason I had it in my head that, because `altis-consent` was loading `altis/consent-api`, that it didn't need to go in here to replace the `wp-consent-api`. 🤦‍♂️

roborourke: Yeah it's not that clever unfortunately, no kind of inheritance logic around that.

PR #94, opened 2020-09-04
Define WP_ENVIRONMENT_TYPE

roborourke:
As part of the WP 5.5 compatibility work this mirrors the `HM_ENV_TYPE` constant to the new core `WP_ENVIRONMENT_TYPE` constant which defaults to "local". This will seemlessly work with the cloud module `wp-config.php` and `wp-config-produciton.php` files should we add `WP_ENVIRONMENT_TYPE` as an environment variable to our infrastructure in future.

See https://github.com/humanmade/altis-cms/issues/239

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -218,9 +218,6 @@ function get_environment_type() : string {
 	if ( defined( 'HM_ENV_TYPE' ) ) {
 		return HM_ENV_TYPE;
 	}
-	if ( defined( 'WP_ENVIRONMENT_TYPE' ) ) {
-		return WP_ENVIRONMENT_TYPE;
-	}
 	return 'local';
 }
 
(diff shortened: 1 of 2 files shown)


inc/namespace.php
```diff
@@ -218,6 +218,9 @@ function get_environment_type() : string {
 	if ( defined( 'HM_ENV_TYPE' ) ) {
 		return HM_ENV_TYPE;
 	}
+	if ( defined( 'WP_ENVIRONMENT_TYPE' ) ) {
+		return WP_ENVIRONMENT_TYPE;
+	}
```
rmccue, line 223: Can we flip this? I'd rather have `WP_ENVIRONMENT_TYPE` set to the output of this function.

inc/namespace.php
```diff
@@ -276,6 +279,22 @@ function fix_plugins_url( string $url, string $path, string $plugin ) : string {
...
+ * Sets the WP_ENVIRONMENT_TYPE constant.
+ *
+ * @return void
+ */
+function set_wp_environment_type() : void {
+	if ( defined( 'WP_ENVIRONMENT_TYPE' ) ) {
+		return;
+	}
+	if ( defined( 'HM_ENV_TYPE' ) ) {
+		define( 'WP_ENVIRONMENT_TYPE', HM_ENV_TYPE );
```
rmccue, line 292: As above.

rmccue (approved):

PR #95, opened 2020-09-15
Fix consent bugs

rmccue:
Fixes up a few issues with the consent PR. Built off #92 to be merged back into that.

### DIFF
--- a/inc/consent/namespace.php
+++ b/inc/consent/namespace.php
@@ -47,9 +47,9 @@ function set_consent_options() {
 	}
 
 	// Check if banner-options was configured.
-	if ( ! empty( $config['banner_options'] ) ) {
+	if ( ! empty( $config['banner-options'] ) ) {
 		// Make sure any option set in the config is a valid option.
-		if ( in_array( $config['banner-options'], Settings\get_cookie_banner_options(), true ) ) {
+		if ( in_array( $config['banner-options'], wp_list_pluck( Settings\get_cookie_banner_options(), 'value' ), true ) ) {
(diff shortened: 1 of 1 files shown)


jazzsequence (approved):

inc/consent/namespace.php
```diff
@@ -108,7 +112,7 @@ function set_analytics_noop( bool $noop ) : bool {
  */
 function set_analytics_data( array $data ) : array {
 	if ( ! wp_has_consent( 'statistics' ) ) {
-		$data['Endpoint']['User'] = [];
+		$data['Endpoint']->User = [];
```
roborourke, line 115: I'll make an issue to better handle this upstream, ideally filters should just be dealing with the array.

inc/consent/namespace.php
```diff
@@ -64,8 +66,10 @@ function set_consent_options() {
 
 	// If any options were set in the config, update those values in the database and remove the options on the settings page.
 	if ( ! empty( $options ) ) {
-		update_option( 'cookie_consent_option', $options );
-		add_filter( 'altis.consent.consent_settings_fields', $fields );
+		add_filter( 'pre_option_cookie_consent_options', function () use ( $options ) {
+			return $options;
+		} );
+		add_filter( 'altis.consent.consent_settings_fields', '__return_empty_array' );
```
roborourke, line 72: Is this right? Some options might be configured but not all I thought

rmccue: Not sure, was just fixing the implementation bug, not necessarily the behaviour. Can be handled in #92 as desired.

PR #107, opened 2020-12-01
update the composer file to add in the consent js api

jazzsequence:
Adds `altis/consent-api` to the composer file so it's not a nested dependency.


roborourke:
Looks like this won't fix the problem. We'd need to split out the composer plugin part of this module into a separate package.

PR #120, opened 2021-01-07
Pull overridden paths dynamically from packages

### LINKED ISSUE #119: Updating Altis requires two calls to `composer update`
Updating Altis requires running `composer update` twice, due to our use of the installer location overrides and the load order of Composer plugins. This subverts user expectations, and often leads to fatal errors which aren't necessarily caught.

Steps to reproduce:
1. Install Altis v4
2. Upgrade to Altis v5 using `composer update`
3. Observe `humanmade/post-cloner` is installed to the wrong location
4. Run `composer update` again
5. Observe `humanmade/post-cloner` is installed to the right location

(Repeat for any major versions as required)

Users should not need to install twice, and files should be placed into the right location after running `composer update` once.

Acceptance criteria:

- [ ] ...

rmccue:
Fixes #119.

Rather than hardcoding the list in altis/core, it pulls from a `extra.altis.install-overrides` entry in each package's `composer.json`. These entries are from the *new* version, rather than the old one, meaning that package entries are always up-to-date.

I've left the legacy list of packages in here until such time as we update every module. We should do this for v6 ideally, then remove the legacy entries in v7 after validating this works.

I've checked this locally briefly, and `$package` is definitely the new copy; additionally, the `extra.altis` entries are there. That said, I want to do more robust testing before we ship anything here.

### DIFF
--- a/inc/composer/class-installer.php
+++ b/inc/composer/class-installer.php
@@ -39,7 +39,7 @@ class Installer extends BaseInstaller {
 	/**
 	 * Gets all overridden packages from all extra.altis.install-overrides entries
 	 *
-	 * @param string[] $overrides
+	 * @param string[] $overrides Overridden package names.
 	 */
 	public function setInstallOverrides( $overrides ) : void {
 		$this->installOverrides = $overrides;
--- a/inc/composer/class-installer.php
+++ b/inc/composer/class-installer.php
@@ -10,6 +10,9 @@ namespace Altis\Composer;
 use Composer\Installers\Installer as BaseInstaller;
 use Composer\Package\PackageInterface;
 
+// phpcs:disable WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase -- Snake case to match Composer.
+// phpcs:disable WordPress.NamingConventions.ValidVariableName.PropertyNotSnakeCase -- Snake case to match Composer.
+
 /**
  * Altis Core Composer Installer.
  */
--- a/inc/composer/class-installer.php
+++ b/inc/composer/class-installer.php
@@ -14,6 +14,7 @@ use Composer\Package\PackageInterface;
  * Altis Core Composer Installer.
  */
 class Installer extends BaseInstaller {
+	protected $installOverrides = [];
 
 	/**
 	 * Check if the installer supports a given type.
@@ -28,24 +29,10 @@ class Installer extends BaseInstaller {
 	/**
 	 * Gets all overridden packages from all extra.altis.install-overrides entries
 	 *
-	 * @return string[]
+	 * @param string[] $overrides
 	 */
-	protected function getAllInstallOverrides() {
-		$rm = $this->composer->getRepositoryManager();
-		$repo = $rm->getLocalRepository();
-		$packages = $repo->getPackages();
-
-		$overridden = [];
-		foreach ( $packages as $package ) {
-			$extra = $package->getExtra();
-			if ( ! isset( $extra['altis'] ) || ! isset( $extra['altis']['install-overrides'] ) ) {
-				continue;
-			}
-
-			$overridden = array_merge( $overridden, $extra['altis']['install-overrides'] );
-		}
-
-		return $overridden;
+	public function setInstallOverrides( $overrides ) : void {
+		$this->installOverrides = $overrides;
 	}
 
 	/**
@@ -98,8 +85,7 @@ class Installer extends BaseInstaller {
 			'stuttter/wp-user-signups',
 		];
 
-		$overrides = $this->getAllInstallOverrides();
-		$excluded_plugins = array_unique( array_merge( $legacy, $overrides ) );
+		$excluded_plugins = array_unique( array_merge( $legacy, $this->installOverrides ) );
 
 		if ( ! in_array( $package->getType(), [ 'wordpress-plugin', 'wordpress-muplugin' ], true ) || ! in_array( $package->getName(), $excluded_plugins, true ) ) {
 			return parent::getInstallPath( $package, $framework_type );
--- a/inc/composer/class-installer.php
+++ b/inc/composer/class-installer.php
@@ -10,12 +10,17 @@ namespace Altis\Composer;
 use Composer\Installers\Installer as BaseInstaller;
 use Composer\Package\PackageInterface;
 
-// @codingStandardsIgnoreStart WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase -- Snake case to match Composer.
-
 /**
  * Altis Core Composer Installer.
  */
 class Installer extends BaseInstaller {
+	/**
+	 * Overridden packages.
+	 *
+	 * Set in setInstallOverrides by the plugin hooks.
+	 *
+	 * @var string[]
+	 */
 	protected $installOverrides = [];
 
 	/**
--- a/inc/composer/class-installer.php
+++ b/inc/composer/class-installer.php
@@ -26,6 +26,29 @@ class Installer extends BaseInstaller {
 	}
 
 	/**
+	 * Gets all overridden packages from all extra.altis.install-overrides entries
+	 *
+	 * @return string[]
+	 */
+	protected function getAllInstallOverrides() {
+		$rm = $this->composer->getRepositoryManager();
+		$repo = $rm->getLocalRepository();
+		$packages = $repo->getPackages();
+
+		$overridden = [];
+		foreach ( $packages as $package ) {
+			$extra = $package->getExtra();
+			if ( ! isset( $extra['altis'] ) || ! isset( $extra['altis']['install-overrides'] ) ) {
+				continue;
+			}
+
+			$overridden = array_merge( $overridden, $extra['altis']['install-overrides'] );
+		}
+
+		return $overridden;
+	}
+
+	/**
 	 * Modifies the install path for Altis module dependencies.
 	 *
 	 * @param PackageInterface $package Composer package manager interface.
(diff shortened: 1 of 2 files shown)


inc/composer/class-installer.php
```diff
@@ -25,6 +25,29 @@ public function supports( $type ) {
 		return in_array( $type, [ 'wordpress-plugin', 'wordpress-muplugin' ], true );
 	}
 
+	/**
+	 * Gets all overridden packages from all extra.altis.install-overrides entries
+	 *
+	 * @return string[]
+	 */
+	protected function getAllInstallOverrides() {
```
joehoyle, line 33: What's the behavior with new installs versus updates? IIRC this won't work for new installs, as at the point of `getInstallPath` being called the packages are not loaded -- that might have just been on the FS though, and I see you are reading it from the `$repo` which maybe is different.

IIRC there are three cases:

1. New install / empty vendor DIR
2. Update with modules all staying the same
3. Update with a module updating, where `$package` _could_ be the old version of the package, not the new one.

joehoyle: Ah re-reading your description, it looks like `$repo->getPackages()` is _always_ the staged/updating versions in all cases?

rmccue: Correct; `$repo` is the local repository, which is the in-memory representation of Composer's state.

Running `composer update` calls our method twice per package: the first time is during a `purgePackages()` call with the existing data, and the second is from [the update itself](https://github.com/composer/composer/blob/2406a094d3b2e5d73ce105764b36898ebd19b69d/src/Composer/Installer.php#L237-L247).

Adding a `var_dump( $package->getName(), $package->getVersion() )` confirms this, with output like:

```
string(9) "altis/seo"
string(7) "4.0.0.0"
...
string(9) "altis/seo"
string(7) "5.0.0.0"
```

Since it's called multiple times, I didn't add any caching in here (which means this method gets called like a hundred times), but I suspect it would be possible to cache it if we listen to the right vars. In particular, I suspect there are changes inside `$repo` we can detect between the first call and the second.

joehoyle: ok nice one, thanks for clarifying

rmccue:
OK, tested with the following process:

* Create installation with v4 using `composer create-project altis/skeleton upgrade-test "~4"`
* Copy the installer file here into vendor to override it
* Add a `var_dump( $overrides )` into the installer file
* Edit composer.json to bump the versions from 4.0.0 to 5.0.0
* Additionally, add `"altis/cms": "dev-install-overrides as 5.0.99"` to incorporate https://github.com/humanmade/altis-cms/pull/300
* Run `composer update`

This produces the following output:
```
$ composer update
array(0) {
}
array(0) {
}
[snip]
Loading composer repositories with package information
Warning from https://repo.packagist.org: You are using an outdated version of Composer. Composer 2.0 is now available and you should upgrade. See https://getcomposer.org/2
Updating dependencies (including require-dev)
Package operations: 1 install, 40 updates, 6 removals
  - Removing symfony/css-selector (v2.8.52)
  - Removing sabberworm/php-css-parser (8.3.1)
  - Removing facebook/graph-sdk (5.7.0)
  - Removing facebook/facebook-instant-articles-sdk-php (v1.10.1)
  - Removing facebook/facebook-instant-articles-sdk-extensions-in-php (v0.2.1)
  - Removing phpunit/phpunit-mock-objects (6.1.2)
  - Updating altis/local-server (4.0.8 => 5.0.2): Loading from cache
  - Updating aws/aws-sdk-php (3.171.13 => 3.171.14): Downloading (100%)
  - Updating altis/core (4.0.3 => 5.0.3): Loading from cache
  - Updating altis/local-chassis (4.0.3 => 5.0.2): Loading from cache
array(0) {
}
array(0) {
}
  - Installing humanmade/post-cloner (1.2.0): Loading from cache
array(0) {
}
  - Updating johnpbloch/wordpress-core (5.4.4 => 5.6.0): Loading from cache
  - Updating johnpbloch/wordpress (5.4.4 => 5.6.0): Loading from cache
  - Removing altis/cms (4.0.7)
  - Installing altis/cms (dev-install-overrides a5b7690): Cloning a5b76905fd from cache
  - Updating symfony/polyfill-mbstring (v1.20.0 => v1.22.0): Downloading (100%)
  - Updating symfony/polyfill-php72 (v1.20.0 => v1.22.0): Downloading (100%)
  - Updating symfony/polyfill-intl-normalizer (v1.20.0 => v1.22.0): Downloading (100%)
  - Updating symfony/polyfill-intl-idn (v1.20.0 => v1.22.0): Downloading (100%)
array(3) {
  [0]=>
  string(21) "humanmade/post-cloner"
  [1]=>
  string(24) "roots/wp-password-bcrypt"
  [2]=>
  string(24) "stuttter/wp-user-signups"
}
array(3) {
  [0]=>
  string(21) "humanmade/post-cloner"
  [1]=>
  string(24) "roots/wp-password-bcrypt"
  [2]=>
  string(24) "stuttter/wp-user-signups"
}
  - Updating altis/aws-analytics (2.2.5 => 2.4.1): Loading from cache
```

So, it is receiving the updated info, but potentially only after the package itself has been updated, which is still too late (as deps are installed before the package is).

Digging some more ♠️

rmccue:
Alright. Cracked it. Following the process above now works; however, note that Composer deprecated the event I had to use in v1, so we have two separate methods for v1 and v2 support here.

I still need to check an initial install with this; Composer v2 is both better and worse at how it handles this, so I'm concerned there might be some sort of issue in there.

rmccue:
OK, so installation doesn't work on v1 or v2 (but, at least it's consistent!) 😬

inc/composer/class-plugin.php
```diff
@@ -51,8 +56,126 @@ public static function getSubscribedEvents() {
...
+	 * @return void
+	 */
+	public function post_dependencies_solving( InstallerEvent $event )  {
+		$operations = $event->getOperations();
+		if ( empty( $operations ) ) {
+			return;
+		}
+
+		// First, work out which packages we already have.
+		$repo = $event->getInstalledRepo();
```
roborourke, line 126: Given this is the only difference between the methods could the hook be derived from the `InstallerEvent` object instead?

rmccue: Possibly; I'd rather have a common method used by both handlers here instead, as I'm wary of heuristically detecting too much. That said, I left that part until I fix the behaviour first

rmccue:
This is now ready for review, and is verified working across the following:

* Composer v1 `update`
* Composer v1 `install`
* Composer v2 `update`
* Composer v2 `install`

To test this out, use `composer create-project altis/skeleton test-project ~4`, then edit your Composer file to upgrade to v5. Additionally, add:
```
        "altis/core": "dev-load-overrides-dynamically as 5.0.99",
        "altis/cms": "dev-install-overrides as 5.0.99"
```

For ease of testing, you may want to take a pristine copy of this directory before you run commands, so that you can test the others easily by simply restoring from this directory.

You should notice that v5 adds `humanmade/post-cloner`, which is correctly installed into `vendor/humanmade/post-cloner`. If you check `vendor/humanmade/core/inc/composer/class-installer.php`, you'll see that post-cloner is not here either before or after the upgrade. Installation into the correct directory indicates success.

----

The implementation here is backwards compatible, so if we would like to, we can backport this. However, we don't need to, as the fix will apply *immediately* upon upgrade; i.e. if we ship this in v6, upgrading to v6 would work seamlessly.

This is late in the day for v6, so if we want to hold off, that'd be fine too.

roborourke:
I'm happy to make an exception for this particular piece! I'm guessing this means we can update the upgrade guide to remove the bit about clearing build cache and removing the `vendor/altis` directory locally?

inc/composer/class-installer.php
```diff
@@ -10,10 +10,13 @@
 use Composer\Installers\Installer as BaseInstaller;
 use Composer\Package\PackageInterface;
 
+// @codingStandardsIgnoreStart WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase -- Snake case to match Composer.
```
roborourke, line 13: I think it's safe to use the `phpcs:ignore` type syntax these days

inc/composer/class-installer.php
```diff
@@ -10,10 +10,13 @@
...
+
 /**
  * Altis Core Composer Installer.
  */
 class Installer extends BaseInstaller {
+	protected $installOverrides = [];
```
roborourke, line 19: Thought the linter would require a docblock but fair enough!

roborourke (approved):
Very minor things in the comments but this is just ace. Sooo glad you figured this out!

rmccue:
>  I'm guessing this means we can update the upgrade guide to remove the bit about clearing build cache and removing the `vendor/altis` directory locally?

Correct.

inc/composer/class-installer.php
rmccue: [I did](https://github.com/humanmade/altis-core/pull/120/commits/3dad9d360ec9b920f55919edb2e0b7397a9dc6d9), and it didn't work, so ¯\_(ツ)_/¯

inc/composer/class-installer.php
rmccue: Hmm, think this might be due to an overzealous ignore then.

inc/composer/class-installer.php
rmccue: Ah, usage of a property and declaration are different 🙄

rmccue:
@roborourke Fixed those up now, I was just blind before 🙈

roborourke:
🚀🚀🚀🚀

PR #126, opened 2021-02-02
Rename installer class to prevent using the old instance

roborourke:
Due to the way composer plugins are evaluated during install we need to ensure the new installer is used and not still in memory by updating the class name. After this change everything works as expected and all plugins are in the correct location.

### DIFF
--- a/inc/composer/class-installer-v6.php
+++ b/inc/composer/class-installer-v6.php
@@ -16,7 +16,7 @@ use Composer\Package\PackageInterface;
 /**
  * Altis Core Composer Installer.
  */
-class Installer extends BaseInstaller {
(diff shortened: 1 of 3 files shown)


jazzsequence:
Will this need to be done every release or is this a Composer v1.1 -> v2 thing?

roborourke:
@jazzsequence just this once, Composer does some janky stuff with eval'ing updated composer plugin code so it got into a situation where it was trying to call a new method on the old instance of the class that was loaded into memory causing a fatal error. Somewhat unrelated, it affects both versions of Composer in reality.

rmccue:
If it's just once, I think we shouldn't tie it to a version number; maybe `OverrideInstaller`?

PR #130, opened 2021-04-30
Add Global Content Repository

roborourke:
This adds the global content repo code and the `wp altis migrate` command necessary to create it.

Related to https://github.com/humanmade/altis-media/issues/100

### DIFF
--- a/docs/global-content-repository.md
+++ b/docs/global-content-repository.md
@@ -8,9 +8,9 @@ By default the Global Content Repository only exposes the user management admin.
 
 Users on the multisite network can have different roles on different sites. This gives you control over who can add and edit global content versus who can only use the global content on their own site.
 
-You might have a user who is an author on the main site, but only a subscriber to the Global Content Repository meaning they could read content to use in their posts but not create or edit global content.
+You might have a user who is an author on the main site, but have no role on the Global Content Repository meaning they could read content to use in their posts but not create or edit global content nor access the global site admin.
 
-Conversely you may have a user who does not have access to your primary site but who can create and edit content on the Global Content Repository.
+Conversely you may have a user who does not have a role on your primary site but who can create and edit content on the Global Content Repository site.
 
 ## Functions
 
@@ -35,6 +35,20 @@ Returns the list of page slugs allowe din the Global Content Repository site adm
 **`altis.core.global_content_site_args : array $args`**
 
 Filters the arguments used to create the Global Content Repository site. The arguments are passed to `wp_insert_site()`.
+
+The below example changes the title, path and domain of the site created. These could also be edited via the network admin as well:
+
+```php
+use Altis\Cloud;
+
+add_filter( 'altis.core.global_content_site_menu_pages', function ( array $args ) : array {
+    // Use the subdomain global.example.org.
+    $args['domain'] = sprintf( 'global.%s', parse_url( Cloud\get_main_site_url(), PHP_URL_HOST ) );
+    $args['path'] = '/';
+    $args['title'] = __( 'Shared Content' );
+    return $args;
+} );
+```
 
 **`altis.core.global_content_site_menu_pages : array $pages`**
 
--- a/docs/global-content-repository.md
+++ b/docs/global-content-repository.md
@@ -12,13 +12,31 @@ You might have a user who is an author on the main site, but only a subscriber t
 
 Conversely you may have a user who does not have access to your primary site but who can create and edit content on the Global Content Repository.
 
+## Functions
+
+**`Altis\Global_Content\get_site_id() : ?int`**
+
+Returns the Global Content Repository site ID or null if it doesn't exist yet.
+
+**`Altis\GlobaL_Content\get_site_url() : ?string`**
+
+Returns the Global Content Repository site URL or null if it doesn't exist yet.
+
+**`Altis\Global_Content\is_global_site( ?int $site_id = null ) : bool`**
+
+Returns `true` if the current site is the Global Content Repository or if the site with the passed `$site_id` is.
+
+**`Altis\Global_Content\get_allowed_admin_pages() : array`**
+
+Returns the list of page slugs allowe din the Global Content Repository site admin menu.
+
 ## Filters
 
-**`altis.core.global_content_site_args: array $args`**
+**`altis.core.global_content_site_args : array $args`**
 
 Filters the arguments used to create the Global Content Repository site. The arguments are passed to `wp_insert_site()`.
 
-**`altis.core.global_content_site_menu_pages: array $pages`**
+**`altis.core.global_content_site_menu_pages : array $pages`**
 
 Filters the allowed top level admin menu pages. Defaults to `[ 'users.php' ]`. To add support for pages you could use the following code:
 
--- a/docs/global-content-repository.md
+++ b/docs/global-content-repository.md
@@ -1,0 +1,31 @@
+# Global Content Repository
+
+Altis includes a Global Content Repository framework. This is a site on the network that can be used to distribute content elsewhere via the REST API or directly to other sites on the network.
+
+By default the Global Content Repository only exposes the user management admin. Other Altis modules may extend this, for example [the Altis Media module has a Global Media Library feature](docs://media/global-media-library.md) built on top of the Global Content Repository that allows for uploading media files and accessing them from all sites on the network.
+
+## User Management
+
+Users on the multisite network can have different roles on different sites. This gives you control over who can add and edit global content versus who can only use the global content on their own site.
+
+You might have a user who is an author on the main site, but only a subscriber to the Global Content Repository meaning they could read content to use in their posts but not create or edit global content.
+
+Conversely you may have a user who does not have access to your primary site but who can create and edit content on the Global Content Repository.
+
+## Filters
+
+**`altis.core.global_content_site_args: array $args`**
+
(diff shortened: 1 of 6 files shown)


inc/global_content/namespace.php
```diff
@@ -0,0 +1,368 @@
...
+	add_filter( 'display_site_states', __NAMESPACE__ . '\\add_global_site_state', 10, 2 );
+}
+
+/**
+ * Get the global content repository site ID.
+ *
+ * @return integer|null
+ */
+function get_site_id() : ?int {
+	return get_site_option( 'global_content_site_id', null );
```
jazzsequence, line 72: This fails to identify if a global content site was initialized but then manually changed to a static url in the config. It also fails to identify manually-set Global Content Site URLs just generally, so a `wp altis migrate` after manually setting the URL to the Global Content Site in the config will create a new `/repo/` site but no site is set as the "global_content_site".

inc/global_content/namespace.php
```diff
@@ -0,0 +1,368 @@
...
+use WP_Site;
+
+/**
+ * Setup global media hooks.
+ *
+ * @return void
+ */
+function bootstrap() : void {
+	// Create the site on upgrade.
+	add_action( 'altis.migrate', __NAMESPACE__ . '\\maybe_create_site' );
```
jazzsequence, line 22: There should be another action triggered at `altis.migrate` that deals with setting a site that's been manually configured as the Global Content Site in the altis config. 

Something like (untested):

```php
add_action( 'altis.migrate', function() {
    $global_media_library_url = Altis\get_config()['media']['global-media-library'];

    // We're running this after the initial `altis.migrate` callback, so if we still don't have a global content site url at this stage, it means that we need to deal with a manually configured site.
    if ( ! get_site_url() && is_string( $global_media_library_url ) ) {
        $domain = stripslashes( parse_url( $global_media_library, PHP_URL_HOST ) );
        $path = stripslashes( parse_url( $global_media_library, PHP_URL_PATH ) );
        $site = get_site_by_path( $domain, $path );

        // Check for an error, e.g. if the site doesn't exist on the network and `get_site_by_path` couldn't find it.
        if ( ! WP_Error( $site ) ) {
   	    update_site_option( 'global_content_site_url', rtrim( $global_media_library_url, '/' ) );
	    update_site_option( 'global_content_site_id', $site->blog_id );
            update_site_meta( $site->blog_id, 'is_global_site', true );
            return;
        }
        // Do some error handling here. Possibly a try / catch would be better than an if / else.
    }
}, 11 );
```

This way the `Global_Content\get_site_id()` and similar functions will still work properly even with manually configured sites.

jazzsequence (changes requested):
I have concerns about manually setting a global content url. Otherwise this looks good.

I also still am wondering about the behavior when I upload media to a non-global site and it still completes the upload but the file disappears when I refresh. But presumably that will get worked out by later issues, so just noting here and will continue to watch it as we build this feature.

jazzsequence:
Here is a gif of the behavior I get when I upload an image from a non-global site currently.

2021-05-03 at 10 55 AM

inc/global_content/namespace.php
roborourke: This is assuming that a manually configured URL is part of the same network which is not the intention. It can be used to connect any other entirely separate WP site as the WP media source.

If the site is part of the same network we should control that process and set up in its entirety.

inc/global_content/namespace.php
roborourke: If it was manually changed the WP AMF source will use that manual setting and it will take precedence. This global content site might still be used by other modules for something like global reusable blocks for example.

inc/global_content/namespace.php
```diff
@@ -0,0 +1,368 @@
...
+		'title' => __( 'Global Content Repository', 'altis' ),
+		'user_id' => get_user_by( 'login', get_super_admins()[0] )->ID,
+		'meta' => [
+			'is_global_site' => true,
+		],
+	] );
+
+	// Ensure user and global site meta is set.
+	if ( ! isset( $global_site_args['user_id'] ) ) {
+		$global_site_args['user_id'] = get_user_by( 'login', get_super_admins()[0] )->ID;
```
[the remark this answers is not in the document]
roborourke: True, I could switch to using `wp_parse_args()` instead 👍 

This value is required for creating a site successfully and you can't trust content that can be filtered, it is possible that the value could be omitted.

docs/global-content-repository.md
```diff
@@ -0,0 +1,49 @@
...
+
+Altis includes a Global Content Repository framework. This is a site on the network that can be used to distribute content elsewhere via the REST API or directly to other sites on the network.
+
+By default the Global Content Repository only exposes the user management admin. Other Altis modules may extend this, for example [the Altis Media module has a Global Media Library feature](docs://media/global-media-library.md) built on top of the Global Content Repository that allows for uploading media files and accessing them from all sites on the network.
+
+## User Management
+
+Users on the multisite network can have different roles on different sites. This gives you control over who can add and edit global content versus who can only use the global content on their own site.
+
+You might have a user who is an author on the main site, but only a subscriber to the Global Content Repository meaning they could read content to use in their posts but not create or edit global content.
```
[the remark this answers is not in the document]
roborourke: They would be able to read media attachments even without being logged in, those are public.

roborourke:
Regarding uploading to the local site I think I linked the pull request in progress on the media module PR but it's being worked on: https://github.com/humanmade/asset-manager-framework/pull/27

It either needs to prevent uploads on the local site right now or support uploading to the global site, I think the PR is working to address the latter.

inc/global_content/namespace.php
jazzsequence: The example code gives preference to sites on the network. If it wasn't on the network a `WP_Error` would be generated and nothing would be changed. In my eyes, if you _can_ manually set a URL to a site on the network, and you want to do that -- rather than using the site that is created automatically -- then the site shouldn't get treated like a second class citizen and not get any of the benefits/enhancements that you would otherwise with the global media site (not unless you don't want the additional restrictions). And if the only difference is setting a meta tag for a different site on the network, we should be able to handle that.

docs/global-content-repository.md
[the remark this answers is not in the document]
jazzsequence: Roles and capabilities are confusing and I don't know that the example helps that much. A user who has a subscriber role on the global content repository site cannot read images without changing the permissions. If this is the intent, we need to build the functionality/change the caps associated with subscribers first. 

Testing locally, I added a subscriber to the global content site and this is what I get when I'm logged in as that user:

Screen Shot 2021-05-04 at 8 38 15 AM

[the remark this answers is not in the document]
roborourke: I was thinking in terms of the REST API here but on the global site admin they wouldn't see the media library so you're right about that. The above would be more accurate if it the user had no role on the global site.

I wasn't thinking about the Editor actually logging into the global site admin in this case.

I need to think on that some more, whether all users should have a role on the global site or just assume or just handle adding it when they do something like upload an image...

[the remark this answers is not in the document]
jazzsequence: Yeah the language makes it sound like an _actual_ subscriber (e.g. user with that role). I will think about it as well and see if I can think of a better way to describe it. 

I think handling when they upload an image probably makes the most sense, otherwise we're potentially migrating a bunch of users that possibly (probably) don't really care.

inc/global_content/namespace.php
roborourke: Consider that you might not want to use the global content repo site as the media library, they don't have to be completely linked. I don't really get why you'd want the `global-media-library` setting from the media module to be able to modify which site is the Global Content Repo site. That's like the tail wagging the dog so to speak.

I added the option to set a URL for the `global-media-library` setting because it's a constant we can define that AMF WP works with so why restrict it?

This would make things more complicated IMO but we could update the docs for the media module to be more explicit about defining a URL being intended for external sites.

I get your intention regarding wanting to be able to define or use another site on the network more easily - so this would be a case of a project defining `AMF_WORDPRESS_URL` on an early hook. We can document how to do that at least but I don't think we should make any further assumptions about their intention for doing that in terms of applying the same modifications as we do to this global content site.

docs/global-content-repository.md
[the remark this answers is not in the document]
roborourke: Updated the language to suggest what happens when there is no role on either site rather than mentioning subscribers. I think as this feature develops we'll get a clearer idea of what we need to document better or build out handling for in the generic sense.

inc/global_content/namespace.php
jazzsequence: this makes sense. I think what needs to be clarified then is the relationship (and distinction) between the Global _Content Site_ and the Global _Media Library_. The two terms are similar and, at least in this phase, are interchangeable, so the assumption is that defining the `global-media-library` in the `composer.json` would set the Global _Content Site_ (even though it doesn't strictly make sense since it's in the Media module, specifically).

roborourke: I'd say even now they aren't interchangeable, I've outlined in the docs in this repo that the Global Media Library is a feature built on top of the Global Content Repository.

I think that assumption might just be a problem for us in this team as that was the original plan, for anyone else we can present it in this new way from the outset. We can put it out to people and see if they make similar assumptions or ask questions along those lines later and make any changes accordingly.

jazzsequence (approved):
👍

PR #139, opened 2021-05-21
remove post cloner

jazzsequence:
This updates the composer installer overrides to remove Post Cloner in preparation of replacing with Yoast Duplicate Post.
We don't need to add Yoast Duplicate Post to replace here because manually overriding in the override installer class is no longer required.

### DIFF
--- a/inc/composer/class-override-installer.php
+++ b/inc/composer/class-override-installer.php
@@ -92,7 +92,6 @@ class Override_Installer extends BaseInstaller {
 			'johnbillion/query-monitor',
 			'stuttter/ludicrousdb',
 			'stuttter/wp-user-signups',
-			'yoast/duplicate-post',
 		];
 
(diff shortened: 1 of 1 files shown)


roborourke:
We don't need these updates any more but makes sense to remove the post cloner item in case it falls back to this list. We should modify the docs around adding dependencies.

jazzsequence:
We don't need to update core for the install path for dependencies anymore?

roborourke:
Yeah that’s what Ryan spent ages sorting back for v6 so we didn’t have this to do this anymore :) Use the `install-overrides` property in the module’s composer file

jazzsequence:
🤯

jazzsequence:
That just made this PR a whole lot simpler 🙄

jazzsequence:
Marking this ready for review now since it's a single line that's not really necessary anymore anyway and doesn't impact anything else.

roborourke (approved):

PR #259, opened 2021-08-18
Pass-through session cookies to global content requests

shadyvb:
This copies and passes-through session cookies to any request to global content site, to allow authenticated requests and support private global content sites.

humanmade/altis-media#159

### DIFF
--- a/inc/global_content/namespace.php
+++ b/inc/global_content/namespace.php
@@ -388,7 +388,7 @@ function filter_global_content_requests_for_auth( array $parsed_args, string $ur
 		return $parsed_args;
 	}
 
-	$parsed_args['cookies'] = array_map( function( $value, $name ) {
+	$parsed_args['cookies'] = array_map( function ( $value, $name ) {
 		return new WP_Http_Cookie( [
 			'name' => $name,
 			'value' => $value,
--- a/inc/global_content/namespace.php
+++ b/inc/global_content/namespace.php
@@ -10,6 +10,7 @@ namespace Altis\Global_Content;
 use Exception;
 use WP_Admin_Bar;
 use WP_CLI;
+use WP_Http_Cookie;
 use WP_Site;
 
 /**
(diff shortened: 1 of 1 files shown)


roborourke (approved):
This is really elegant, nice work

inc/global_content/namespace.php
```diff
@@ -366,3 +370,30 @@ function prevent_site_deletion( array $caps, string $cap, int $user_id, array $a
...
+ * @return array Parsed request arguments including current session cookies.
+ */
+function filter_global_content_requests_for_auth( array $parsed_args, string $url ) : array {
+	$global_site_url = get_site_url();
+
+	if ( strpos( $url, $global_site_url ) !== 0 ) {
+		return $parsed_args;
+	}
+
+	$parsed_args['cookies'] = array_map( function( $value, $name ) {
```
rmccue, line 391:
```suggestion
	$parsed_args['cookies'] = array_map( function ( $value, $name ) {
```

:D

inc/global_content/namespace.php
```diff
@@ -61,6 +62,9 @@ function bootstrap_site() : void {
 	// Handle network admin sites list.
 	add_filter( 'manage_sites_action_links', __NAMESPACE__ . '\\sites_list_row_actions', 10, 2 );
 	add_filter( 'display_site_states', __NAMESPACE__ . '\\add_global_site_state', 10, 2 );
+
+	// Handles passing current user cookies to global content requests to allow authenticated requests.
+	add_filter( 'http_request_args', __NAMESPACE__ . '\\filter_global_content_requests_for_auth', 10, 2 );
```
rmccue, line 67: Applying this to all requests might cause some weird unintentional behaviour; can we add it to individual requests instead, or is that going to be a gigantic pain?

shadyvb: Not really, Rob asked for a global solution so it works with shared blocks and other features in the future.

We can always whitelist requests if needed. 

Unless we have a specific concern in mind about this, I'd say let's keep it and address issues as they come up.

I think the more concerning issue would be the selection of cookies to send, rather than the requests to handle. But again I couldn't think of a reason not to forward everything.

roborourke: Yeah I'd rather this solution "just work" for anyone building on top of the global content repo. This is only going to apply to requests coming from within the same application fwiw.

I agree with Shadi in terms of iterating on it if we do find issues, possibly around user roles and permissions for individuals that differ between sites on the network.

One of my previous attempts to solve this involved automatically adding users making requests to the global site to it with the contributor role by default fwiw.

inc/global_content/namespace.php
roborourke: Surprised linter bot didn't catch that 🤔

shadyvb: Was just about to say the same.

roborourke:
As a follow up we should also default the global content repo site to being private.

PR #314, opened 2021-10-18
Update module registration to pass module options

roborourke:
Part of humanmade/product-dev#367


shadyvb (approved):
LGTM, one comment though is that we use `settings` and `options` interchangeably here, would've been nice to consolidate.

PR #360, opened 2021-11-15
Add codeception tests for Release QA tasks

shadyvb:
Fixes humanmade/product-dev#838

### DIFF
--- a/tests/acceptance/AdminCest.php
+++ b/tests/acceptance/AdminCest.php
@@ -1,0 +1,49 @@
+<?php // phpcs:disable
+/**
+ * Tests for module version checks
+ */
+
+/**
+ * Test for module version checks
+ */
+class AdminCest {
+
+	public function moduleVersionsDisplayed( AcceptanceTester $I ) {
+		$I->wantToTest( 'About page shows module versions.' );
+		$I->loginAsAdmin();
+		$I->amOnAdminPage( 'about.php' );
+		$I->see( 'Current Module Versions' );
+		$modules = [
+			'CMS',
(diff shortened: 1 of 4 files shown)


tests/integration/global-content-repo/GlobalContentRepoTest.php
```diff
@@ -0,0 +1,19 @@
+<?php // phpcs:disable
```
roborourke, line 1: I think it's worth following our coding standards even in the tests, I know the generator outputs the stub differently but there's not loads to edit to bring things in line.

tests/acceptance/AdminCest.php
```diff
@@ -0,0 +1,33 @@
...
+			'Media',
+			'Multilingual',
+			'Privacy',
+			'Security',
+			'SEO',
+			'SSO',
+			'Workflow',
+		];
+		foreach ( $modules as $module ) {
+			$I->see( $module );
```
roborourke, line 29: This isn't checking versions right now. Maybe not the most valuable of tests so we can move on if you prefer but would be good to verify the version. Should be accessible from the `composer.lock` or `vendor/composer/installed.json`

roborourke (commented):
Couple of suggestions. Should we remove the `\Helper` class references if they're not used right now?

It'd be good to try and follow our coding standards too, I'm not sure what CC's defaults / expectations are around class and method naming though. Not a blocker but would be good to keep the codebase as consistent as possible.

roborourke (approved):
Awesome work

PR #371, opened 2022-01-14
Add Segment.io Telemetry

roborourke:
Porting Accelerate plugin telemetry system to core.

Related to https://github.com/humanmade/product-dev/issues/874

Screenshot 2022-01-14 at 16 30 44

### DIFF
--- a/inc/telemetry/namespace.php
+++ b/inc/telemetry/namespace.php
@@ -1,0 +1,326 @@
+<?php
+/**
+ * Altis Product Telemetry.
+ *
+ * @package altis/core
+ */
+
+namespace Altis\Telemetry;
+
+use Altis;
+use Segment\Segment;
+use WP_REST_Request;
+use WP_REST_Server;
+use WP_User;
+
+const API_NAMESPACE = 'altis/v1';
+const ACTION_OPT_IN = 'altis_tracking_opt_in';
+const META_OPT_IN = 'altis_tracking_opt_in';
+const SEGMENT_ID = 'GHqd7Vfs060yZBWOEGV4ajz3S3QHYKhk';
+
+/**
+ * Register hooks.
+ *
+ * @return void
+ */
+function bootstrap() {
+	add_action( 'admin_head', __NAMESPACE__ . '\\load_segment_js' );
+	add_action( 'admin_footer', __NAMESPACE__ . '\\render_identity_tag' );
+
+	add_action( 'rest_api_init', __NAMESPACE__ . '\\register_api_routes' );
+
+	// Default event tracking.
+	add_action( 'save_post', __NAMESPACE__ . '\\track_new_or_updated_content', 10, 3 );
+
+	// Allow action hook for tracking, this makes it easy to track events in other code
+	// without having a direct dependency on this module.
+	add_action( 'altis.telemetry.track', __NAMESPACE__ . '\\track' );
+}
+
+/**
+ * Initialize segment.io.
+ */
+function initialize() {
+	static $initialized;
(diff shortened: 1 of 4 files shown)


roborourke:
Actually we do have some UI stuff in this module already so I'll add the basic opt in UI here too.

roborourke:
It's basic styling but we can remove it in the analytics plugin in favour of the React UI.

inc/telemetry/namespace.php
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	$id = get_network_option( null, 'altis_tracking_id', null );
+	if ( ! empty( $id ) ) {
+		return $id;
+	}
+
+	// Hash the registration date of the first site to get a unique ID.
+	$site = get_blog_details( get_main_site_id(), false );
+	$generated = 'local-' . substr( hash( 'sha1', $site->registered ), 0, 6 );
```
tfrommen, line 194: This might be a bit edge-case, but it's possible that two or more sites have the same registration timestamp, if done programmatically. Why not use something that is actually required to be unique, like the ID full domain? 🤔 (Maybe I don't understand how/what this is used for, or why the date is being used to generate the hash.)

rmccue: These are user IDs across development machines, so ID = 1 and domain is going to overlap across all devs on the same project generally.

inc/telemetry/namespace.php
```diff
@@ -0,0 +1,445 @@
...
+ * @return void
+ */
+function load_segment_js() {
+	?>
+	<script>
+		// Segment - Load
+		!function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on","addSourceMiddleware","addIntegrationMiddleware","setAnonymousId","addDestinationMiddleware"];analytics.factory=function(e){return function(){var t=Array.prototype.slice.call(arguments);t.unshift(e);analytics.push(t);return analytics}};for(var e=0;e<analytics.methods.length;e++){var key=analytics.methods[e];analytics[key]=analytics.factory(key)}analytics.load=function(key,e){var t=document.createElement("script");t.type="text/javascript";t.async=!0;t.src="https://cdn.segment.com/analytics.js/v1/" + key + "/analytics.min.js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(t,n);analytics._loadOptions=e};analytics._writeKey="GHqd7Vfs060yZBWOEGV4ajz3S3QHYKhk";analytics.SNIPPET_VERSION="4.13.2";
```
tfrommen, line 234: Does this change (often)? Would it be better moved to an actual JavaScript file?

rmccue (approved):
Looks good to me but I feel like I'm partially just reviewing my own code so you might want a second review?

inc/telemetry/namespace.php
tfrommen: Aha, I see. You want to distinguish multiple local development environments...?

inc/telemetry/namespace.php
rmccue: This is Segment's snippet which we use verbatim from them.

roborourke: Well potentially we could punt it to tag manager instead but for the time being this is fine, changing it is pretty easy and backporting if needed is easy too.

PR #374, opened 2022-01-19
Ensure segment init has run

roborourke:
If posts are created or updated programmatically with no logged in user context the initialize function exists early and `Segment::track()` is called without `Segment::init()` getting run.

### DIFF
--- a/inc/telemetry/namespace.php
+++ b/inc/telemetry/namespace.php
@@ -43,15 +43,18 @@ function bootstrap() {
 
 /**
  * Initialize segment.io.
+ *
+ * @return bool True if Segment has initialised successfully.
  */
-function initialize() {
+function initialize() : bool {
 	static $initialized;
-	if ( $initialized ) {
-		return;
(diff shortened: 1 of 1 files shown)


inc/telemetry/namespace.php
```diff
@@ -43,15 +43,18 @@ function bootstrap() {
...
-	if ( $initialized ) {
-		return;
+	if ( is_bool( $initialized ) ) {
+		return $initialized;
 	}
 
 	if ( ! is_user_logged_in() ) {
-		return;
+		$initialized = false;
+		return $initialized;
```
robindevitt, line 57: Any particular reason to not use `return false;`

robindevitt (approved):

inc/telemetry/namespace.php
roborourke: Not really, I guess just so it's completely clear we're always returning whatever initialised gets set to, even though that should technically be obvious.

PR #378, opened 2022-01-27
Use line for output in CLI

roborourke:
`WP_CLI::log()` doesnt show the output during the installation process.


mikelittle (commented):
Hmmm... I'm not sure what this is addressing. But deliberately bypassing the quiet flag seems... unusual. If the output from `log()` is not seen is that because it is being called with the --quiet flag? ( in which case the solution maybe to not call with that flag) or because the output is being redirected to a file or other process? (in which case the solution may be to pipe the output through `tee` first so it can go to the file/process and stdout). See https://www.geeksforgeeks.org/tee-command-linux-example/

PR #398, opened 2022-02-11
Display upgrade notice even on new dash

### LINKED ISSUE #397: Upgrade notice no longer shows with new insights dashboard
As of Altis v10 the new insights dashboard enabled by default replaces the standard dashboard where we show the upgrade notice for Altis versions out of Long Term Support.

We need to update to an approach that will work regardless of whether the dashboard is overridden at all. I propose a modal using the WP Core UI elements (see media modal for example markup).

Acceptance criteria:

- [ ] Modal should have the same content as current upgrade notice
- [ ] Modal should only show on the dashboard / home page of the admin
- [ ] Modal should be dismissable and not return
- [ ] Modal should only be shown to super admins

rmccue:
Fixes #397.

New Dash | Network Dash (or Old Dash)
-- | --
Screenshot 2022-02-11 at 12-14-06 Home ‹ Altis — Altis | Screenshot 2022-02-11 at 12-14-43 Dashboard ‹ Network Admin Altis Sites — Altis

It's a little ugly on the old dash, but to fix that we'd need a dependency between the analytics and the core module. IMO, it's acceptable.

### DIFF
--- a/inc/upgrades/namespace.php
+++ b/inc/upgrades/namespace.php
@@ -232,12 +232,19 @@ function render_widget() {
 				>
 					<?php
 						// phpcs:ignore
-						echo clean_html( sprintf(
-							'%1$s <span class="screen-reader-text">%2$s</span>',
-							__( 'Learn more about Altis upgrades' ),
-							/* translators: Accessibility text. */
-							__( '(opens in a new tab)' )
-						), 'span' );
+						echo clean_html(
+							sprintf(
+								'%1$s <span class="screen-reader-text">%2$s</span>',
+								__( 'Learn more about Altis upgrades' ),
+								/* translators: Accessibility text. */
+								__( '(opens in a new tab)' )
+							),
+							[
+								'span' => [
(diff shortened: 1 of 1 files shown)


inc/upgrades/namespace.php
```diff
@@ -13,14 +13,10 @@
...
-	add_action( 'wp_network_dashboard_setup', __NAMESPACE__ . '\\register_widget' );
-	add_action( 'wp_dashboard_setup', __NAMESPACE__ . '\\register_widget' );
+	add_action( 'load-index.php', __NAMESPACE__ . '\\maybe_render_header', 0 );
 }
 
-/**
- * Register our dashboard widget.
- */
-function register_widget() {
+function maybe_render_header() {
```
roborourke, line 19:
```suggestion
/**
 * Conditionally render the upgrade notice.
 */
function maybe_render_header() {
```

roborourke (approved):
Missing docblock but added a suggestion

PR #402, opened 2022-02-16
Add upgrade prompts when not on the latest version

rmccue:
Fixes https://github.com/humanmade/product-dev/issues/914


roborourke:
I don't love it personally, but if it's an exec decision the only thing I'd suggest is maybe check with marketing if they want any campaign params for reporting on those links.

shadyvb (approved):

Issue #418, opened 2022-02-28
Global Content Repository urls


robindevitt:
Inform users that when using content from the Global Content Repository, that the URL's will be relevant to the Global Content Repository and not the site it's being displayed on.

PR #419, opened 2022-02-28
Global Content Repository urls

robindevitt:
Added a brief note that informs users that when using content from the Global Content Repository, that the URL's will be relevant to the Global Content Repository and not the site it's being displayed on.

linked issue #418

### DIFF
--- a/docs/global-content-repository.md
+++ b/docs/global-content-repository.md
@@ -6,7 +6,7 @@ By default the Global Content Repository only exposes the user management admin.
 
 The Global Content Repository site is set to be a private site by default. This can be modified by switching the site to a public one from the Network Admin > Sites management page, clicking `Edit` on the site, and checking the `Public` checkbox.
 
-**Note:**When using media from the Global Content Repository, the URL for the selected media will always be relative to that of your Global Content Repository site and not the site it's being displayed on.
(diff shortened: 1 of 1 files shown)


docs/global-content-repository.md
```diff
@@ -6,6 +6,8 @@ By default the Global Content Repository only exposes the user management admin.
 
 The Global Content Repository site is set to be a private site by default. This can be modified by switching the site to a public one from the Network Admin > Sites management page, clicking `Edit` on the site, and checking the `Public` checkbox.
 
+When using media from the Global Content Repository, the url for the selected media will always be relevant to that of your Global Content Repository site and not the site it's being displayed on.
```
roborourke, line 9: I usually like to call something like this out with a bit of bold text to start like:
```suggestion
**Note:** When using media from the Global Content Repository, the URL for the selected media will always be relative to that of your Global Content Repository site and not the site it's being displayed on.
```
IMO it's a good habit for making it more likely people will see the gotchas when looking at docs.

Also changed:
- relevant -> relative (when talking about URLs they are relative or absolute, relevant kind of means the same but doesn't make sense in this context)
- url -> URL (acronyms should be capitalised)

roborourke (approved):
One suggestion to commit then good to go. Pre-approving so as not to hold you up.

docs/global-content-repository.md
robindevitt: Thanks for the feedback Rob.

PR #424, opened 2022-03-02
Correctly hash for automatic IDs

### LINKED ISSUE #423: Automatic IDs in Segment are incorrect
Steps to reproduce:
1. Opt out of personal data collection in Telemetry
2. Observe ID is 32303231

We're using `bin2hex` and then truncating this value, however the value is a datetime so we're just getting a hex representation of `2022` (or `2021`). 🤦 

Acceptance criteria:

- [ ] Different anonymous users have unique IDs

rmccue:
Fixes #423


roborourke (approved):
I think we should change the use of bin2hex below as well. A hash is a much better option as a long email address = a longer hex string.

Also why do we take a substring of the auto ID, is that necessary?

roborourke (approved):

PR #442, opened 2022-03-30
Fix Composer v2 operation type usage

### LINKED ISSUE #441: Installation broken with Composer 2.3.0
When running Composer 2.3.0 released today the Altis installation breaks with the following message:

```
In class-plugin.php line 124:
                                                                                              
  [Error]                                                                                     
  Cannot access protected const Composer\DependencyResolver\Operation\InstallOperation::TYPE 
```

We need to fix this asap as Travis builds are all using the latest version and breaking.

There seem to be problems with some other projects too:

- https://github.com/composer/composer/issues/10672
- https://github.com/composer/composer/issues/10671

Seldaek:
Fixes #441

### DIFF
--- a/inc/composer/class-plugin.php
+++ b/inc/composer/class-plugin.php
@@ -121,7 +121,7 @@ class Plugin implements PluginInterface, EventSubscriberInterface {
 	 */
 	protected function resolve_packages_composer_v2( array $packages, array $operations ) : array {
 		foreach ( $operations as $operation ) {
-			switch ( $operation::TYPE ) {
+			switch ( $operation->getOperationType() ) {
 				case 'install':
 				case 'markAliasInstalled':
 					$package = $operation->getPackage();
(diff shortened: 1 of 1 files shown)


Seldaek:
I did not check anything sorry just trying to help out here. getOperationType exists as of Composer 2.0.0, so probably you should rather bump composer-plugin-api to ^2.0 as well. Unless Composer 1 usage is still needed, then you could also call getJobType if getOperationType method does not exist for BC?

roborourke:
It's really appreciated, just checking now and it does seem to work but trying to make sense of where that method comes from. Can't seem to find reference to it in the Composer codebase

Seldaek:
Fixed now so it supports Composer 1 as well.

roborourke:
That's my fault for getting confused, we are already differentiating between v1 and v2 before hitting that method so `getOperationType()` is all we need - I'll do the update and handle the rest, thanks so much for your time and help, you're an absolute legend.

roborourke (approved):

Seldaek:
Ah ok great :)

PR #465, opened 2022-05-04
Check `get_blog_details()` exists for local ID

roborourke:
Fixes a fatal if site is a multisite but ms-blogs functions have not yet been included.


roborourke:
Plugin was using an outdated version of Altis core, should be resolved now.

PR #490, opened 2022-06-12
Add missing dependency

szepeviktor:
Used in inc/upgrades/namespace.php five times.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,7 @@
 		"aws/aws-sdk-php": "^3.224.3",
 		"composer-plugin-api": "^1.1 || ^2.0",
 		"composer/installers": "^1.12 || ^2.0",
+		"humanmade/clean-html": "^2.0",
 		"segmentio/analytics-php": "~2.0.0"
 	},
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -14,6 +14,7 @@
 		"aws/aws-sdk-php": "^3.224.3",
 		"composer-plugin-api": "^1.1 || ^2.0",
 		"composer/installers": "^1.12 || ^2.0",
+		"humanmade/clean-html": "^2.0",
```
roborourke, line 17:
```suggestion
		"humanmade/clean-html": "^2.0.0",
```

roborourke (commented):
It's a dependency of `altis/cms` currently. Are you using a subset of Altis modules?

szepeviktor:
> Are you using a subset of Altis modules?

I'm not a user. This is a random contribution.

roborourke:
Ok cool, well thank you!

mikelittle:
This is already included indirectly by composer dependencies. No need to include it directly. Thanks.

PR #506, opened 2022-07-01
Content changes to track in Altis

robindevitt:
[Content changes to track in Altis](https://github.com/humanmade/product-dev/issues/1040)

### DIFF
--- a/inc/telemetry/namespace.php
+++ b/inc/telemetry/namespace.php
@@ -511,17 +511,58 @@ function track_new_or_updated_content( $post_id, $post, $update ) {
 		return;
 	}
 
+	// Set the event value.
 	if ( $update ) {
 		$action = 'update';
 	} else {
 		$action = 'create';
 	}
 
+	// Set the visibility value.
+	if ( 'private' === $post->post_status ) {
+		$visibility = 'Private';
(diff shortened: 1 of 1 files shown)


inc/telemetry/namespace.php
```diff
@@ -507,21 +511,93 @@ function handle_telemetry_endpoint( WP_REST_Request $request ) {
  */
 function track_new_or_updated_content( $post_id, $post, $update ) {
 	// Bail on auto-save.
-	if ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) {
+	if ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE || 'auto-draft' === $post->post_status ) {
 		return;
 	}
```
robindevitt, line 516: I have had to include the `auto-draft` status here in the return as it seems to apply when a new post ( the exact post_type ) is created but not other post types. Not returning then fires off the tracking event. Essentially firing two tracking events when creating a new post.

roborourke: I would do this in a separate if statement. Read through the logic here, I don't think it will do what you in every scenario. What if `DOING_AUTOSAVE` is defined but false. You should never mix `&&` operators directly with `||`.

PR #513, opened 2022-07-11
Update Segment tracking with the page title.

robindevitt:
Issue - [Update Segment Tracking for page view event in Altis #1038](https://github.com/humanmade/product-dev/issues/1038)

This PR adds the admin page title to the default Segment.io tracking.

By default `get_admin_page_title()` returns **Home** for the **Dashboard** and **Welcome** for the **Developer Documentation**  page.

I have included a switch statement to return **Dashboard** and **Developer Documentation**  for the relevant pages.

### DIFF
--- a/inc/telemetry/namespace.php
+++ b/inc/telemetry/namespace.php
@@ -299,12 +299,23 @@ function get_environment_details() : array {
  * @return void
  */
 function load_segment_js() {
+	$page_name = get_admin_page_title();
+	switch ( $page_name ) {
+		case 'Home':
+			$page_name = 'Dashboard';
+			break;
+		case 'Welcome':
+			$page_name = 'Developer Documentation';
+			break;
+		default:
+			$page_name;
+	}
 	?>
 	<script>
 		// Segment - Load
 		!function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on","addSourceMiddleware","addIntegrationMiddleware","setAnonymousId","addDestinationMiddleware"];analytics.factory=function(e){return function(){var t=Array.prototype.slice.call(arguments);t.unshift(e);analytics.push(t);return analytics}};for(var e=0;e<analytics.methods.length;e++){var key=analytics.methods[e];analytics[key]=analytics.factory(key)}analytics.load=function(key,e){var t=document.createElement("script");t.type="text/javascript";t.async=!0;t.src="https://cdn.segment.com/analytics.js/v1/" + key + "/analytics.min.js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(t,n);analytics._loadOptions=e};analytics._writeKey="<?php echo esc_js( get_segment_id() ); ?>";analytics.SNIPPET_VERSION="4.13.2";
 			analytics.load( <?php echo wp_json_encode( get_segment_id() ) ?> );
-			analytics.page();
+			analytics.page( "<?php echo $page_name ;?>" );
 		}}();
 	</script>
 	<?php
--- a/inc/telemetry/namespace.php
+++ b/inc/telemetry/namespace.php
@@ -299,16 +299,10 @@ function get_environment_details() : array {
  * @return void
  */
 function load_segment_js() {
-	$page_name = get_admin_page_title();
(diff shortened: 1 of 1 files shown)


veselala:
@robindevitt that's a good call, thank you!

roborourke:
It would be ideal for the developer documentation to be able to set the admin page title depending on the current documentation page - can you make an issue for that?

inc/telemetry/namespace.php
```diff
@@ -299,12 +299,23 @@ function get_environment_details() : array {
  * @return void
  */
 function load_segment_js() {
+	$page_name = get_admin_page_title();
+	switch ( $page_name ) {
+		case 'Home':
+			$page_name = 'Dashboard';
```
roborourke, line 305: Are the admin page title names translatable? This might not match if they are.

Additionally does 'Dashboard' refer to our replacement dashboard / accelerate? I would avoid having this kind of logic here ideally and aim to modify the admin page title in the context of the features that change it. I'm presuming `get_admin_page_title()` can be filtered?

Rather than looking at the page title string to decide whether to modify it we should look at the conditions for whether it should be modified, for instance on the docs pages `$_GET['page']` will equal `altis-documentation`: https://github.com/humanmade/altis-documentation/blob/master/inc/ui/namespace.php#L14-L46

roborourke (changes requested):

robindevitt:
>It would be ideal for the developer documentation to be able to set the admin page title depending on the current documentation page - can you make an issue for that?

I looked into this some more. It appears that the when adding the submenu pages the titles are set. 
https://github.com/humanmade/altis-documentation/blob/master/inc/ui/namespace.php#L50-L62

I did a quick look through and it appears that when viewing the developer docs the value for `get_admin_page_title();` is that set by the `h1` title tag of the respective doc. So when viewing the "Developer Documentation" initially, the `h1` title is "Welcome", which makes sense as to where that's coming from. I have confirmed this is the case for when viewing other documents within the developer docs.

inc/telemetry/namespace.php
robindevitt: It looks to be that both the dashboard and accelerate when viewing the `/wp-admin` page returns as "Home".

`get_admin_page_title()` looks like it can be filtered, https://developer.wordpress.org/reference/hooks/admin_title/, should we wish for the "Home" value to be changed still.

inc/telemetry/namespace.php
```diff
@@ -299,12 +299,17 @@ function get_environment_details() : array {
  * @return void
  */
 function load_segment_js() {
+	if ( isset( $_GET['page'] ) ){
+		$page_name = str_replace( '-', ' ', ucfirst( $_GET['page'] ) );
```
roborourke, line 303: I think there has to be a way to get the page title passed through in the `add_menu_page()` / `add_menu_subpage(0` functions.. can you look into that? Maybe using `get_current_screen()` and the `WP_Screen` object?

I'm never keen on approaches like this using functions like `ucfirst` and `ucwords`, they're almost never appropriate if there's a genuine human readable string available somewhere.

roborourke (changes requested):
You need to fix the linter errors, added a comment too - I think this can be improved a bit.

inc/telemetry/namespace.php
roborourke: Yeah that's not surprising. So in the analytics plugin we could modify the page title using that filter if we're showing the accelerate dashboard.

inc/telemetry/namespace.php
robindevitt: I haven't found an alternative way to get the the page title with your suggestions.

What if we stuck with using `get_admin_page_title();` for the title, then use the the filter `admin_title` filter to change the title as and when needed?

Each of the pages in the documents when converting the markdown and an `h1` tag is set, it then overrides the original page title.

roborourke: > What if we stuck with using get_admin_page_title(); for the title, then use the the filter admin_title filter to change the title as and when needed?

Yep agreed, that's the approach I'm after here. The only special case we have is our replacement dashboard - I'll make a PR to filter that and tag you for review. Then the implementation here can just be to use `get_admin_page_title()`.

Noticed a general issue with the admin docs page titles that it'd be good to fix to make this more useful https://github.com/humanmade/altis-documentation/issues/397

roborourke:
PR for dashboard page title here https://github.com/humanmade/aws-analytics/pull/448

inc/telemetry/namespace.php
roborourke: So that filter and the function are a little strange - the filter modifies the entire title string, including the site name, while the function sets the default value for the `$title` global variable, unless one has already been set.

roborourke (approved):

PR #519, opened 2022-07-19
Fix parsing of installed.json for Composer 1.x

kovshenin:
Before 2.x Composer's installed.json looked slightly different. Tests in this module are currently failing with Composer 1.x because of the difference in format. This commit adds support for both versions.

Here's the current output from Travis under Composer 1.x:

```
1) AdminCest: Test About page shows correct module versions.
 Test  [secure]/core/tests/acceptance/AdminCest.php:moduleVersionsDisplayed
                                                                                    
  [PHPUnit\Framework\Error\Notice] Trying to get property 'packages' of non-object  
                                                                                    
Scenario Steps:
 3. $I->see("Current Module Versions") at [secure]/core/tests/acceptance/AdminCest.php:26
 2. $I->amOnAdminPage("about.php") at [secure]/core/tests/acceptance/AdminCest.php:25
 1. $I->loginAsAdmin() at [secure]/core/tests/acceptance/AdminCest.php:24
#1  /usr/src/app/vendor/[secure]/core/tests/acceptance/AdminCest.php:29
#2  AdminCest->moduleVersionsDisplayed
```

### DIFF
--- a/tests/acceptance/AdminCest.php
+++ b/tests/acceptance/AdminCest.php
@@ -17,6 +17,8 @@ class AdminCest {
 	 *
 	 * @param AcceptanceTester $I Tester
 	 *
+	 * @throws \Exception if Composer's installed.json file could not be parsed.
+	 *
 	 * @return void
 	 */
 	public function moduleVersionsDisplayed( AcceptanceTester $I ) {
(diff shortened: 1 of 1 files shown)


shadyvb (approved):
TIL!

PR #521, opened 2022-07-20
Update dependabot.yml for v12

kovshenin:
Adds v12-branch, removes v8-branch.


roborourke (approved):
HM Linter error is unrelated, and I think new

PR #571, opened 2022-09-07
Show the error in WP-CLI when creating the global media site fails

johnbillion:
WP-CLI doesn't catch and display the message from an exception, so `WP_CLI::error()` needs to be called explicitly. This was causing an error from the global media site creation to not appear in the output when updating Altis with `composer server cli -- altis migrate`.

## Before:

```
Success: WordPress database upgraded on 2/2 sites.
Creating Global Content Repository site...
```

## After:

```
Success: WordPress database upgraded on 2/2 sites.
Creating Global Content Repository site...
Error: Global media site could not be created. Sorry, that site already exists!
```


roborourke (approved):
Thanks @johnbillion !

Issue #453, opened 2022-04-11
Upgrade notice shows when using a pre-release e.g. beta version


roborourke:
When running Altis v11 beta 1 I see the upgrade notice in the admin telling me to update to v10.

Acceptance criteria:

- [ ] Upgrade notice does not appear when current is greater than current stable version

PR #604, opened 2022-10-10
Don't show the upgrade notice for beta versions.

robindevitt:
Related issue - #453 

Currently when on a beta version of Altis, the upgrade notice is shown. 

This PR ensures that when your "current" version of Altis, is equal to or greater than the "latest" version you won't be shown the upgrade notice.

### DIFF
--- a/inc/upgrades/namespace.php
+++ b/inc/upgrades/namespace.php
@@ -129,7 +129,7 @@ function is_version_latest() : bool {
 	}
 
 	$releases = get_supported_version_info();
-	return $version === ( $releases[0]['version'] ?? $version );
+	return ( $version >= $releases[0]['version'] ?? false );
 }
 
 /**
--- a/inc/upgrades/namespace.php
+++ b/inc/upgrades/namespace.php
@@ -129,7 +129,7 @@ function is_version_latest() : bool {
 	}
 
 	$releases = get_supported_version_info();
-	return ( $version >= $releases[0]['version'] ?? false );
+	return $version >= ( $releases[0]['version'] ?? $version );
 }
 
 /**
--- a/inc/upgrades/namespace.php
+++ b/inc/upgrades/namespace.php
@@ -352,13 +352,6 @@ function render_upgrade_widget() {
 	}
 
 	$latest = $releases[0];
-	$current = get_current_version_info();
-
-	// Return early for versions greater or equal to the latest version.
-	if ( $current['version'] >= $latest['version'] ) {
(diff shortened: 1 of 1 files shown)


inc/upgrades/namespace.php
```diff
@@ -352,6 +352,13 @@ function render_upgrade_widget() {
 	}
 
 	$latest = $releases[0];
+	$current = get_current_version_info();
+
+	// Return early for versions greater or equal to the latest version.
+	if ( $current['version'] >= $latest['version'] ) {
```
rmccue, line 358: I think we should fix the check inside `is_version_latest()` instead

robindevitt: Thanks for the suggestion Ryan, definitely a better approach!

inc/upgrades/namespace.php
```diff
@@ -129,7 +129,7 @@ function is_version_latest() : bool {
 	}
 
 	$releases = get_supported_version_info();
-	return $version === ( $releases[0]['version'] ?? $version );
+	return ( $version >= $releases[0]['version'] ?? false );
```
robindevitt, line 132: Would this be preferred?
`return ( isset( $releases[0]['version'] ) && $version >= $releases[0]['version'] ?? false );`

rmccue: `??` is the [null coalescing operator](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op) which replaces the need to use isset.

That said, the intention here is to handle if there's no value available to force the check to return true, so your code needs to be
```suggestion
	return $version >= ( $releases[0]['version'] ?? $version );
```

rmccue (changes requested):

rmccue:
I can't review this seeing as I wrote the suggestion :D

kovshenin (approved):
Looks good!

robindevitt:
@rmccue I can't merge this until you approve the requested changes too, please. I tried to remove you but was only able to re-request a review. @kovshenin has approved the change too, your approval should allow the merge.

kovshenin:
@robindevitt you're good to go, I removed the blocker 😆

robindevitt:
@kovshenin is this worth me backporting to v13 & v12? Maybe include v11?

kovshenin:
This only affects future betas. The first v13 beta has already shipped, and we're not planning a second one currently, so I don't think it needs backporting.

PR #632, opened 2022-11-18
Pin aws/aws-sdk-php to allow patch increments only

kovshenin:
A change in the AWS SDK version 3.243.0 introduced a new class to construct dynamic endpoints, which caused problems with S3 Uploads. Let's pin the SDK version to an earlier version and allow patch increments only, and have Dependabot handle the rest.

See:

* https://github.com/aws/aws-sdk-php/issues/2563
* https://github.com/humanmade/altis-local-server/issues/542

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
 	],
 	"require": {
 		"php": ">=7.1",
-		"aws/aws-sdk-php": "^3.242.1",
+		"aws/aws-sdk-php": "~3.242.1",
 		"composer-plugin-api": "^1.1 || ^2.0",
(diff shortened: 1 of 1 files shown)


roborourke:
Wondered if that would ever bite us. That's annoying. Note that dependabot will still try to update to a newer minor version until you use the `@dependabot ignore this minor version` command on the next PR it makes.

We can maybe assess whether to remove the GH action I added to auto merge the dependabot PRs on this repo once tests passed. I never got around to applying it elsewhere.

roborourke (approved):

PR #692, opened 2023-02-14
Update dependabot for vX release

mikelittle:


mikelittle:
Bad commit name (copy pasta error)

PR #736, opened 2023-04-20
Introduce a JSON schema for composer.json

johnbillion:
It's possible to use `allOf` in JSON Schema to combine the constraints from multiple schemas into one. This means we can create a schema for composer.json which is a combination of [the official Composer schema](https://getcomposer.org/schema.json) with our own schema for the `extra.altis` properties.

Adding `"$schema": "vendor/altis/core/schema.json",` to the top of an application-level composer.json file means VS Code provides all the JSON schema goodness you'd expect for a complete schema, including autocomplete, hover documentation with links, and validation.

By using a root `definitions` we also allow all the module definitions to appear in `extra.altis.modules` as well as each environment within `extra.altis.environments` without duplicating everything five times.

## Screenshots

Top level modules:

Environment modules:

Autocomplete

## Status

This is a proof of concept. There is a maintenance overhead for such a schema so we could make more use of `$ref` in order to pull in each module definition from its place in the vendor directory if that's preferable.

Thoughts?

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -1,0 +1,149 @@
+{
+	"$schema": "http://json-schema.org/draft-07/schema#",
+	"$id": "https://schemas.altis-dxp.com/schema.json",
+	"title": "Altis Core schema for composer.json",
+	"type": "object",
+	"definitions": {
+		"altis-modules": {
+			"title": "Configuration for Altis modules",
+			"description": "https://docs.altis-dxp.com/getting-started/configuration/",
+			"type": "object",
+			"properties": {
+				"analytics": {
+					"title": "Analytics module for Altis",
+					"description": "https://docs.altis-dxp.com/analytics/",
+					"type": "object",
+					"properties": {
+						"enabled": {
+							"description": "Whether this module is enabled",
+							"type": "boolean",
+							"default": true
+						},
+						"native": {
+							"description": "Whether to use native analytics",
+							"type": "boolean",
+							"default": true
+						}
+					}
+				},
+				"cms": {
+					"title": "CMS module for Altis",
+					"description": "https://docs.altis-dxp.com/cms/",
+					"type": "object",
+					"properties": {
+						"enabled": {
+							"description": "Whether this module is enabled",
+							"type": "boolean",
+							"default": true
+						},
+						"xmlrpc": {
+							"description": "Whether the XML-RPC API is enabled",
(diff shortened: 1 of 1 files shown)


rmccue:
> Yeah we can do this by putting a schema.json into each module which contains its module definition, and then the main schema.json in `altis/core` would pull each of them in via `"$ref": "../analytics/schema.json"`.

Right, but that then centralises a dependency in `altis/core` on each module. We could have that live in `altis/altis` if we have to (rather than core) but that's kind of what I mean

johnbillion:
I think centralising the _list of modules_ for this purpose is reasonable and manageable. As far as I know we don't have a high turnover of new or removed modules. The alternative would be to come up with a build step that combines the schemas from all the modules, but I'm not too keen on that, and presumably that would also require centralising a list of the modules to combine.

The schemas for the modules themselves would live within each module so maintaining the module schemas would remain the responsibility of each module.

> We could have that live in altis/altis if we have to

Yeah I used `altis/core` only because `altis/altis` is a meta package and doesn't result in anything being present in `vendor/altis/altis` for an application. With `altis/core` we already have the `vendor/altis/core` directory in which to place the schema.

rmccue:
> As far as I know we don't have a high turnover of new or removed modules

That is true, but it also breaks the conceptual modularity by having dependencies between the modules, so we want to avoid it as much as possible.

Additionally, we support both custom modules and a mixed list of modules depending on which plan a user is on; the latter is a minimal problem for something like this was the worst case is you get too many suggestions, but still avoidable I think.

> The alternative would be to come up with a build step that combines the schemas from all the modules, but I'm not too keen on that, and presumably that would also require centralising a list of the modules to combine.

This can be decentralised using the same mechanism we use for installation path overrides, where modules could specify a `extra.altis.config-schema` in their `composer.json`, and we'd build all of that together; see also, [how we do install overrides](https://github.com/humanmade/altis-core/blob/0f0db4596ee8632fd75a12645d2089e0914a965d/inc/composer/class-plugin.php#L262-L274).

That's definitely my preferred option, provided the schema will work correctly with a local file rather than a full URL, which it seems is perfectly fine.

joehoyle:
As a start, could we atleast define the `altis` config block and `modules` / `environments` keys. Those are not specific to any modules, and then we could add module-specific config (hopefully pulling the schema from each module).

PR #779, opened 2023-07-19
Add custom configuration for Action Scheduler

### LINKED ISSUE #780: WooCommerce jobs eat up PHP workers unnecessarily
If the background processing in Cavalcade has a minor blip, or jobs start to exceed typical times, Action Scheduler (WooCommerce's custom job processing) will skip over using cron and start using async HTTP requests. As this happens when the queue processing is starting to be slow, the async processing can _also_ be slow.

Worse, because these are then routed to the web workers rather than Cavalcade, autoscaling of Cavalcade won't kick in correctly, and PHP workers will be used up unnecessarily causing the web containers to be misused.

We can tweak Action Scheduler's behaviour to ensure it always uses wp-cron, and then further optimise how it uses wp-cron to ensure efficient processing. This should improve AS performance as well as avoiding impacts on the web containers from backlogged queues.

Acceptance criteria:

- [ ] When WooCommerce is activated, Action Scheduler does not spawn async HTTP requests to process jobs

rmccue:
Optimises Action Scheduler to always use Cavalcade, instead of falling back to async requests. In addition, tweaks some numbers to ensure queues are processed efficiently.

Fixes #780

See also, https://github.com/woocommerce/action-scheduler/blob/trunk/docs/perf.md

@joehoyle to tweak and test these numbers further.

### DIFF
--- a/inc/compatibility/namespace.php
+++ b/inc/compatibility/namespace.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Altis\Compatibility;
+
+/**
+ * Bootstrap compatibility tweaks.
+ */
+function bootstrap() {
+	add_action( 'plugins_loaded', __NAMESPACE__ . '\\set_woocommerce_compatibility' );
+}
+
+/**
+ * Set hooks for WooCommerce compatibility and optimization.
+ *
+ * Adjusts Action Scheduler to run optimially with Cavalcade, by avoiding
+ * fallbacks and maximising the Cavalcade queue.
+ *
+ * The time limit and batch size hooks are set at priority 0 to allow sites to
+ * easily override for their specific use case.
+ */
+function set_woocommerce_compatibility() {
+	if ( class_exists( '\\ActionScheduler' ) ) {
+		return;
+	}
+
+	// Disable Action Scheduler's async request fallback to ensure it runs
+	// via Cavalcade.
+	add_filter( 'action_scheduler_allow_async_request_runner', '__return_false', 1000 );
+
+	// Allow Action Scheduler to run for 5 minutes (Cavalcade's limit is 60,
+	// but most hosts stay around 1-2 minutes)
+	add_filter( 'action_scheduler_queue_runner_time_limit', function () {
+		return 5 * MINUTE_IN_SECONDS;
+	}, 0 );
+
+	// Increase queue batch size from 25 to 250: we give 12x the default time
+	// (5 minutes instead of 30s), so should be able to run 8x the number of jobs comfortably.
+	add_filter( 'action_scheduler_queue_runner_batch_size', function () {
+		return 250;
+	}, 0 );
+}
--- a/inc/compatibility/namespace.php
+++ b/inc/compatibility/namespace.php
@@ -19,7 +19,7 @@ function bootstrap() {
  * easily override for their specific use case.
  */
 function set_woocommerce_compatibility() {
-	if ( class_exists( '\\ActionScheduler' ) ) {
+	if ( ! class_exists( '\\ActionScheduler' ) ) {
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


joehoyle:
We pushed this to one customer, and things look to be operating as expected. I don't have data on the exact numbers, but it seems to be ok as-is.

rmccue:
Is it better than before?

johnbillion:
I'd love to see the reason for the changes documented more clearly. On the rs-dms project we tweaked these settings quite a lot and made sure to document the changes clearly.

See:

*https://github.com/humanmade/rs-dms/blob/master/client-mu-plugins/rs-bulk-management/inc/functions.php#L74-L79
* https://github.com/humanmade/rs-dms/blob/master/client-mu-plugins/rs-bulk-management/inc/scheduler.php

joehoyle:
For documentation, I think we could elaborate on the in-code docs in https://github.com/humanmade/altis-core/pull/779/files#diff-66101e5c9a2ddcfa6f9bfa5ec8c36099c49038ceffc4182eede19db6f9406990R17-R25. @johnbillion is that along the lines you're thinking?

joehoyle:
> Is it better than before?

The issue we're seeing in this case is that the fallback exists at all. The numbers on how many jobs to do at once etc is not really that important. The larger issue is that Action Scheduler will by default start making very slow HTTP requests (blocking PHP workers) if the cron jobs get slightly backed up. This PR changes that behavior so it's more binary in that sense. Yes, it's better, because we now don't use the fallback behavior and block the PHP workers doing so.

inc/compatibility/namespace.php
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/**
+ * Set hooks for WooCommerce compatibility and optimization.
+ *
+ * Adjusts Action Scheduler to run optimially with Cavalcade, by avoiding
+ * fallbacks and maximising the Cavalcade queue.
+ *
+ * The time limit and batch size hooks are set at priority 0 to allow sites to
+ * easily override for their specific use case.
```
joehoyle, line 24:
```suggestion
 * easily override for their specific use case.
 *
 * Specifically we disable the action_scheduler_allow_async_request_runner. This is functionality in
 * Action Scheduler that will fall back to making async requests to admin-ajax.php to process the
 * queued jobs. This will typically get triggered if the Cron is expected not to be working. The issue we've
 * seen with this approach is when Cavalcade has a somewhat long queue, Action Scheduler switches to the 
 * fallback behavior, which we don't want. We don't want that because it eats up available PHP workers, where
 * we prefer to just wait for Cavalcade to get to the queued jobs.
 *
 * We're also adjusted the limits on how many Action Scheduler Jobs will run in a single Action Schedule
 * Cron Job as it's more efficient to run that way when we know how long Cavalcade jobs can run as opposed
 * to the default WordPress Cron runner which likely has lower timeouts.
```

joehoyle:
@rmccue I added the above suggestion for document this further.

inc/compatibility/namespace.php
rmccue: This is preeeeettttyyyy wordy, and kinda repeats what the code says; how about:

```suggestion
 * easily override for their specific use case.
 *
 * We disable the async task runner entirely, to avoid the loopback HTTP
 * requests to admin-ajax.php. This triggers when AS detects cron not working,
 * but doesn't account for the Cavalcade backlog and scaling system, and
 * triggers too eagerly. Instead, we prefer a slight delay as Cavalcade scales
 * up, avoiding the loopbacks.
 *
 * The limits are adjusted to account for the better processing power and
 * limits of Cavalcade.
```

svandragt:
Hey is there anything that can be done to progress this @rmccue ? We're seeing a lot of missed action scheduler quee items (couple dozen)

joehoyle:
@rmccue added your comment suggestion. Can you re-review please.

joehoyle:
Oh wait you opened the PR

joehoyle (approved):

PR #840, opened 2023-10-16
Remove the legacy package list for install path overrides

### LINKED ISSUE #839: Remove legacy install path overrides for old packages
With removal of some modules and potentially more in future there is a list of legacy packages that Altis will override the installation path for. This should be seen as a bug because it's unexpected and hard to debug behaviour.

https://github.com/humanmade/altis-core/blob/master/inc/composer/class-override-installer.php#L55-L92

The list should be updated and backported, there may be some other packages to remove depending on the Altis version.

roborourke:
The legacy list was for Composer v1 and the transition between the 2. I think we can ditch this to avoid errors in future where packages are unintentionally forced into the vendor directory.

Fixes #839


mikelittle (approved):
Tested locally. all good.

PR #1047, opened 2024-05-27
Add API to get secrets/variables

### LINKED ISSUE #1041: Add API to retrieve variables and secrets
Per [the docs](https://github.com/humanmade/altis-cloud/pull/878), this should be:

```php
/**
 * Get the value of a variable or secret.
 *
 * @param string $name Name of the variable or secret. Must match the name in the Altis Dashboard exactly.
 * @param mixed $default Default value to return if not set (such as in local environments).
 * @return string|mixed Value as a string if set, default value otherwise.
 */
Altis\get_variable( string $name, mixed $default = null ) : mixed;
```

rmccue:
Fixes #1041.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -585,3 +585,23 @@ function powered_by( $format = null ) {
 		)
 	);
 }
+
+/**
+ * Get the value of a variable or secret.
+ *
+ * This includes both secrets and non-secret variables.
+ *
+ * @param string $name Name of the variable or secret. Must match the name in the Altis Dashboard exactly.
+ * @param mixed $default Default value to return if not set (such as in local environments).
+ * @return string|null Value as a string if set, null otherwise.
+ */
+function get_variable( string $name ) : ?string {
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -585,3 +585,23 @@ function powered_by( $format = null ) {
...
 	);
 }
+
+/**
+ * Get the value of a variable or secret.
+ *
+ * This includes both secrets and non-secret variables.
+ *
+ * @param string $name Name of the variable or secret. Must match the name in the Altis Dashboard exactly.
+ * @param mixed $default Default value to return if not set (such as in local environments).
```
mikelittle, line 595: This `@param` doc needs removing now the parameter has been removed.
```suggestion
```

mikelittle (changes requested):
Just a doc block change to match the param removal.

kovshenin (approved):

mikelittle (approved):

PR #1127, opened 2024-08-29
Update composer installers to accept 2.3.0

mikelittle:
This is a manual application of a too-old Dependabot PR


wisyhambolu (approved):
LGTM

Issue #1298, opened 2025-09-29
Build wp altis post-sync (or similar) CLI command


VlasicVojislav:

jerico:
As I understand this, we'll just make an action hook available (e.g. `altis.post-sync`) that we will execute via `wp altis post-sync` at the end of the sync process.

VlasicVojislav:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Altis-6284e0dfb386b00c83981426/issues/gh/humanmade/altis-core/1298?planning-poker) with Zenhub @ferschubert-hm @jerico @mikelittle @vladislavhmn @wisyhambolu

PR #1405, opened 2026-04-10
Add `wp altis post-sync` CLI command

mikelittle:
## Summary

Introduces the `wp altis post-sync` CLI command and the `altis.post_sync` action hook. Custom code and other Altis modules can hook into this command to run maintenance tasks after an environment sync (for example, pulling a production database to staging or development).

This PR ships one default handler in core:

- **Truncate Cavalcade logs** — truncates the `cavalcade_logs` table (skips if the table is not found).

The handler is a named function (`Altis\Post_Sync\truncate_cavalcade_logs`) that customers can unhook individually via `remove_action()`.

Other modules contribute their own handlers in their own packages. In particular, the Elasticsearch reindex handler that previously lived here has moved to the Search module — see humanmade/altis-enhanced-search#573.

- Addresses #1298

## Test plan

- [ ] Run `wp altis post-sync` — completes successfully and the Cavalcade logs table is truncated.
- [ ] On a site without the `cavalcade_logs` table, verify the truncate step is skipped with a log message rather than erroring.
- [ ] Verify a custom action can be hooked: `add_action( 'altis.post_sync', function () { WP_CLI::log( 'Custom task ran' ); } );`.
- [ ] Verify the default action can be removed: `remove_action( 'altis.post_sync', 'Altis\\Post_Sync\\truncate_cavalcade_logs' );`.
- [ ] Check `wp help altis post-sync` shows the command documentation.
- [ ] With humanmade/altis-enhanced-search#573 merged/checked out locally, verify the Elasticsearch reindex still runs when `wp altis post-sync` is invoked.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/cli-command.md
+++ b/docs/cli-command.md
@@ -41,8 +41,9 @@ WP_CLI::add_hook( 'after_invoke:altis migrate', function () {
 The post-sync command runs routine maintenance tasks after syncing an environment, for example after pulling a production 
 database to staging or development. By default, this includes:
 
-- Reindexing Elasticsearch (if available)
 - Truncating the Cavalcade cron logs table
+- Reindexing Elasticsearch when the Search module is enabled (see
+  [Search module docs](docs://search/cli-command.md))
 
 ### Hooking into the command
 
@@ -69,12 +70,12 @@ There are 2 ways to hook into this command:
 All default tasks are registered as named functions, so they can be unhooked individually:
 
 ```php
-// Skip the Elasticsearch reindex.
-remove_action( 'altis.post_sync', 'Altis\Post_Sync\reindex_elasticsearch' );
-
 // Skip truncating Cavalcade logs.
 remove_action( 'altis.post_sync', 'Altis\Post_Sync\truncate_cavalcade_logs' );
 ```
+
+Tasks contributed by other modules (for example the Search module's Elasticsearch reindex) are documented
+alongside those modules.
 
 ### Customer examples
 
--- a/docs/cli-command.md
+++ b/docs/cli-command.md
@@ -41,7 +41,6 @@ WP_CLI::add_hook( 'after_invoke:altis migrate', function () {
 The post-sync command runs routine maintenance tasks after syncing an environment, for example after pulling a production 
 database to staging or development. By default, this includes:
 
-- Flushing the object cache
 - Reindexing Elasticsearch (if available)
 - Truncating the Cavalcade cron logs table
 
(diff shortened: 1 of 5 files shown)


jerico:
My understanding is  `Flush object cache is hooked to the action`, `Reindex elasticsearch is hooked to the action` , and `Truncate cavalcade logs in the DB is hooked to the action` are meant to be examples of what customers can do with post-sync e.g. provide in documentation how to do it. Not make it default actions.

This is based on:
- We always cache flush after database import https://github.com/humanmade/vantage-backend/pull/2404#discussion_r3018030090
- Not all environment have elasticsearch enabled 
- Cavalcade logs are customer data and it's their responsibility/call if they want to clear or truncate it

@joehoyle @rmccue

mikelittle:
> My understanding is `Flush object cache is hooked to the action`, `Reindex elasticsearch is hooked to the action` , and `Truncate cavalcade logs in the DB is hooked to the action` are meant to be examples of what customers can do with post-sync e.g. provide in documentation how to do it. Not make it default actions.
> 

I didn't get that from the ticket. But if we are already flushing the cache, I'll remove that part.
It does check whether ES is available, so doesn't expect everyone will have it.

I'll wait for more feedback from Joe and Ryan to figure out what to do.

If all the quote above is true, this should not have been 5 story points.

inc/post_sync/namespace.php
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Reindex Elasticsearch if available.
+ *
+ * Skips reindexing if Elasticsearch is not configured for this environment.
+ *
+ * @return void
+ */
+function reindex_elasticsearch() : void {
+	if ( ! defined( 'ELASTICSEARCH_HOST' ) || ! ELASTICSEARCH_HOST ) {
```
joehoyle: Would this not be better in the search module?

rmccue:
Huh, thought I'd submitted a comment here but guess not? Sorry for the delay!

I did mean those actions specifically rather than just as an example, as I think they're basically needed every time you import data. Sounds like cache flushing is already handled in Vantage though, and agreed that we should have the ES one as a hook in the search module.

jerico:
Glad its cleared out. Looks good once ES is moved to the search module as Joe suggested

docs/cli-command.md
```diff
@@ -35,3 +35,73 @@ WP_CLI::add_hook( 'after_invoke:altis migrate', function () {
     WP_CLI::runcommand( 'cli info' );
 } );
 ```
+
+## `wp altis post-sync`
```
jerico, line 39: Could this be in it's own standalone page? Concerned that it's buried under "Altis CLI Command".

We also link to the docs within the sync modal more straightforward to link directly to standalone page.

Whichever path we go, just need to update this https://github.com/humanmade/vantage/pull/1053 when the final permalink is known.

mikelittle: I think it fits best within the page it is. Both commands are similar and fit together.

inc/post_sync/namespace.php
mikelittle: Yep. Make sense. Moved. New PR is https://github.com/humanmade/altis-enhanced-search/pull/573

joehoyle (approved):

